PR #5, opened 2020-05-20
Base server implementation with REST endpoints

danielSanchezQ:
First draft with the chain data endpoint loading data from static json file.


danielSanchezQ:
This code may be change but it establishes the intention of the project organization.

PR #13, opened 2020-05-28
Proposals REST endpoint load from db

danielSanchezQ:


src/v0/endpoints/proposals/logic.rs
```diff
@@ -1,4 +1,22 @@
...
+use diesel::RunQueryDsl;
+
+pub async fn get_all_proposals(context: SharedContext) -> Vec<Proposal> {
+    let db_conn = context
+        .read()
+        .await
+        .db_connection_pool
+        .get()
+        .expect("Error connecting to database");
+    tokio::task::spawn_blocking(move || {
```
danielSanchezQ, line 12: I tried to create a function to abstract this call but couldn't make it work. I have a question in the Diesel gitter. Hopefully they answer and I can do it. For now it should be enough anyway.

danielSanchezQ: From the `diesel` gitter chat:

> You will need a lot more trait bounds to make something like that work. I normally do advice not to go down that route.

No harmful of having it duplicated.

PR #18, opened 2020-06-08
Mount playground endpoint

### LINKED ISSUE #11: Expose a graphiql endpoint
We should be able to expose a graphiql endpoint for testing. It is available in the [`async-graphql` library](https://docs.rs/async-graphql/1.13.2/async_graphql/http/fn.graphiql_source.html) that we are using already.

This endpoint should be exposed based on [configuration](https://github.com/input-output-hk/vit-servicing-station/issues/10).

danielSanchezQ:
Playground is the next iteration of the graphiql endpoint. It is just exposed when building a debug binary.
Closes #11

### DIFF
--- a/src/v0/endpoints/graphql/graphiql.rs
+++ b/src/v0/endpoints/graphql/graphiql.rs
@@ -1,0 +1,14 @@
+use warp::{Filter, Rejection, Reply};
+
+#[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


danielSanchezQ:
> 
> 
> Why only on debug build?

Not anymore :)

PR #21, opened 2020-06-09
Graphql id queries

danielSanchezQ:
- Implemented fund by id query
- Implement proposal by id query
- Properly handle errors on queries


rinor:
LGTM

PR #22, opened 2020-06-13
API tokens manager

danielSanchezQ:
Created a manager and utilities around it for abstracting the handling of API tokens

### DIFF
--- a/src/db/models/api_token.rs
+++ b/src/db/models/api_token.rs
@@ -4,7 +4,7 @@ use diesel::Queryable;
 #[derive(Queryable)]
 #[diesel(table = "api_tokens")]
 pub struct APIToken {
-    token: String,
+    token: Vec<u8>,
     creation_time: String,
     expire_time: String,
 }
--- a/src/db/models/api_token.rs
+++ b/src/db/models/api_token.rs
@@ -1,10 +1,22 @@
+use crate::db::{schema::api_tokens, DB};
+use crate::v0::api_token::APIToken;
 use diesel::Queryable;
 
 #[allow(dead_code)]
-#[derive(Queryable)]
-#[diesel(table = "api_tokens")]
-pub struct APIToken {
-    token: Vec<u8>,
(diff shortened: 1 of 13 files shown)


src/db/models/api_token.rs
```diff
@@ -0,0 +1,10 @@
+use diesel::Queryable;
+
+#[allow(dead_code)]
+#[derive(Queryable)]
+#[diesel(table = "api_tokens")]
+pub struct APIToken {
+    token: String,
+    creation_time: String,
+    expire_time: String,
```
NicolasDP, line 9: why are we using `String`? Why not `[u8; 32]`, `u64` and `u64` ?

src/v0/api_token.rs
```diff
@@ -0,0 +1,81 @@
...
+use diesel::query_dsl::{filter_dsl::FilterDsl, RunQueryDsl};
+use diesel::{ExpressionMethods, OptionalExtension};
+use warp::{Filter, Rejection};
+
+/// Header where token should be present in requests
+const API_TOKEN_HEADER: &str = "API-Token";
+
+/// API Token wrapper type
+#[derive(PartialEq, Eq)]
+pub struct APIToken(String);
```
NicolasDP, line 16: For the API Token, us a `[u8; 32]`. We will figure out the crypto for this later. In the header, indeed, encode it in base64 URL

src/db/models/api_token.rs
danielSanchezQ: Yes, the token should be some kind of hash. For the dates, it is actually to keep coherence with the other types we are currently using (encoded as strings).

danielSanchezQ: Also for the current backend having dates as strings makes easy to make some queries:
https://sqlite.org/lang_datefunc.html

danielSanchezQ:
I'm merging this, it works. More changes are coming in the next iterations and they will make an impact here too.

PR #25, opened 2020-06-23
Serialization deserialization adjustments and endpoints testing

danielSanchezQ:
- Added aliases for graphql responses objects deserializing
- Implemented some testing utilities
- Implemented some serialization/deserialization helper methods
- Implemented tests for REST and GraphQL endpoints


danielSanchezQ:
Going ahead with this. More changes incoming relying on this.

PR #26, opened 2020-06-25
Refactored all common db queries into a queries module

danielSanchezQ:


danielSanchezQ:
Going ahead with this refactor. It can be rethink later if see something does not fit.

PR #30, opened 2020-06-30
[Tests] Vit station sanity test

dkijania:
Added first sanity test. This PR is also to present approach in integration tests.
What is important :
- split main into lib and main to be able to import crate as lib and use it in integration tests (it's impossible for cli only)
- added testing module, where is put all test utils. This will make easier to maintain code
- added `test-api` feature 
- added few test dependencies
- added conditional compilation for testing module
- added integration test

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ async-graphql = "1.13.2"
 async-graphql-warp = "1.13.2"
 async-trait = "0.1.33"
 base64 = "0.12.1"
+cfg-if = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 diesel = { version = "1.4.4", features = ["sqlite", "r2d2"] }
 dotenv = "0.9.0"
@@ -22,12 +23,29 @@ structopt = "0.3.14"
 thiserror = "1.0"
 tokio = { version = "0.2.21", features = ["macros"] }
 warp = { version = "0.2", features = ["tls"] }
+diesel_migrations = "1.4.0"
+#test dependencies
+
+#diesel doesn't have an way to create empty db
+#rustqlite is used for that purpose
+rusqlite = "0.14"
+tempfile = "3"
+assert_fs = "1.0.0"
+assert_cmd = "1.0.1"
+[dependencies.reqwest]
+version = "0.10.6"
+default-features = false
+features = ["blocking"]
 
 # This solves building on windows when sqlite3lib is not installed or missing in the `$PATH`
 # as it happens with the github actions pipeline associated to this project.
 [target.'cfg(windows)'.dependencies]
 libsqlite3-sys =  { version = "0.9.1", features = ["bundled"] }
 
+[features]
+test-api = []
+
 [dev-dependencies]
-diesel_migrations = "1.4.0"
-tempfile = "3"
+vit-servicing-station = {path= ".", features= ["test-api"] }
+
+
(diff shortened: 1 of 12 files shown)


src/testing/startup/db.rs
```diff
@@ -0,0 +1,108 @@
...
+        if let Some(migrations_folder) = &self.migrations_folder {
+            self.do_migration(&connection, migrations_folder)?;
+        }
+        Ok(())
+    }
+
+    fn try_insert_token(&self, connection: &SqliteConnection) -> Result<(), DbBuilderError> {
+        if let Some(token) = &self.token {
+            let values = (
+                api_tokens::dsl::token.eq(&(*token.token.as_ref())),
```
danielSanchezQ, line 62: I don’t understand this referencing/ de referencing dance. Nitpicking, but, can’t it be simplified?

Cargo.toml
```diff
@@ -22,12 +23,29 @@ structopt = "0.3.14"
 thiserror = "1.0"
 tokio = { version = "0.2.21", features = ["macros"] }
 warp = { version = "0.2", features = ["tls"] }
+diesel_migrations = "1.4.0"
+#test dependencies
+
+#diesel doesn't have an way to create empty db
```
danielSanchezQ, line 29: You mean to create a new db file?

src/testing/mod.rs
```diff
@@ -0,0 +1,20 @@
+pub mod rest_client;
+pub mod server;
+pub mod startup;
+
+use crate::{db::models::api_tokens::APITokenData, v0::api_token::APIToken};
+
+use chrono::offset::Utc;
+
+pub fn get_testing_token() -> (APITokenData, String) {
```
danielSanchezQ, line 9: It’s ok for now here. But I tried to keep the Sam structure in the project so it i easy to  locate everything. Maybe we can have a sub module for this kind of calls (testing-data?). Inside a token file where this would be. Even if we make it public from the mod later, so we keep it easy to locate and offer a simple way and an explicit way of importing it. What do you think?

src/testing/rest_client.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum RestError {
+    #[error("could not deserialize response")]
+    CannotDeserialize(#[from] serde_json::Error),
+    #[error("could not send reqeuest")]
+    RequestError(#[from] reqwest::Error),
+}
```
danielSanchezQ, line 55: Same way we can structure this: a mod clients, one for rest and later a graphql will be coming probably.

Cargo.toml
dkijania: yes, i couldn't find any way using diesel, to create new db. while rustqlite `Connection::open` method can create new db if the file it pointing does not exist

src/testing/startup/db.rs
dkijania: it was `cargo clippy` suggestion. I will take another look

src/testing/mod.rs
dkijania: agree. I've put it here because it didn't want to make module for only one method. But  this is a good idea to have separate module for test data

src/testing/rest_client.rs
dkijania: also agree. Will change that

src/testing/startup/db.rs
danielSanchezQ: I think `api_tokens::dsl::token.eq(token.token.as_ref())` should be enough.

dkijania:
Closing PR as it easier to create new one after breaking changes made in project

PR #35, opened 2020-07-02
Divide project in sub crates

danielSanchezQ:
- Refactor project structure
- Update async-graphql deps to latest version (1.16.2)

The idea is that when growing the project it can be more maintainable and easier to structure. This change was wanted before starting implementing different testing vectors.


danielSanchezQ:
Testing went well after refactor and adjustments. Going ahead with this PR

PR #38, opened 2020-07-03
Skip api token filter for playground

### LINKED ISSUE #31: remove api token requirement when using graphql playground endpoint
It's particularly hard to use graph ql playground from browser when user is required to send header with api-token.

danielSanchezQ:
Fix #31


dkijania:
now playground is launched correctly, but there is an error "Server cannot be reached" while trying to use it

danielSanchezQ:
Apparently that is due to some `CORS` settings. I will have to take a look on how to solve that exactly. Could't find any specific solution but it is a commonly error in the graphql playground implementations.

dkijania:
ok, I will alter the settings and recheck

danielSanchezQ:
@dkijania , the thing is to access it through `127.0.0.1` instead of `localhost`: 

`http://127.0.0.1:3030/api/v0/graphql/playground`

I'm not sure why it doesn't work if accessing from `localhost` (I guess at some point it is not recognized). I'm gonna open a question in warp.

dkijania:
@danielSanchezQ i was using 127.0.0.1, got the same error like you

danielSanchezQ:
But could you make a request? There are 2 errors here. The CORS one that is solved using `127.0.0.0` instead of `localhost` and an error that is the same as #28, but queries still work.

danielSanchezQ:
It works for me. Using the new `async-graphql` version `OutputJson` solves  #28 too. Playground retrieves schema and doc succesfully now.

PR #40, opened 2020-07-06
[Tests] Added test project

dkijania:
Added new project for testing with two sanity tests for rest call and graphql call

### DIFF
--- a/vit-servicing-station-tests/src/common/data/mod.rs
+++ b/vit-servicing-station-tests/src/common/data/mod.rs
@@ -1,0 +1,60 @@
+use chrono::offset::Utc;
+use vit_servicing_station_lib::{
+    db::models::{
+        api_tokens::APITokenData,
+        proposals::{Category, Proposal, Proposer},
+        vote_options::VoteOptions,
+    },
+    v0::api_token::APIToken,
+};
+
+pub struct TestGen;
+
+impl TestGen {
+    pub fn token() -> (APITokenData, String) {
+        let data = b"ffffffffffffffffffffffffffffffff".to_vec();
+        let token_data = APITokenData {
+            token: APIToken::new(data.clone()),
+            creation_time: Utc::now().timestamp(),
+            expire_time: Utc::now().timestamp(),
+        };
+        (
+            token_data,
+            base64::encode_config(data, base64::URL_SAFE_NO_PAD),
+        )
+    }
+
+    pub fn proposals() -> Vec<Proposal> {
+        vec![Proposal {
+            internal_id: 1,
+            proposal_id: "1".to_string(),
+            proposal_category: Category {
+                category_id: "".to_string(),
+                category_name: "foo_category_name".to_string(),
+                category_description: "".to_string(),
+            },
+            proposal_title: "the proposal".to_string(),
+            proposal_summary: "the proposal summary".to_string(),
+            proposal_problem: "the proposal problem".to_string(),
+            proposal_solution: "the proposal solution".to_string(),
+            proposal_public_key: "pubkey".to_string(),
+            proposal_funds: 10000,
+            proposal_url: "http://foo.bar".to_string(),
+            proposal_files_url: "http://foo.bar/files".to_string(),
+            proposer: Proposer {
+                proposer_name: "tester".to_string(),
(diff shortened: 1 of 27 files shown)


vit-servicing-station-tests/src/tests/graphql.rs
```diff
@@ -0,0 +1,9 @@
+use crate::common::startup::quick_start;
+use assert_fs::TempDir;
+
+#[test]
+pub fn graphql_sanity_test() {
```
danielSanchezQ, line 5: I think it would be better to be explicit about what we are testing. In this case, If we get a proposal by id lets reflect it in the name somehow. Or am I missing the point of the test here? Is this gonna run all functionality in one test?

vit-servicing-station-tests/src/common/startup/db.rs
```diff
@@ -0,0 +1,179 @@
...
+            proposals: None,
+        }
+    }
+
+    pub fn with_token(&mut self, token: APITokenData) -> &mut Self {
+        self.token = Some(token);
+        self
+    }
+
+    pub fn with_proposals(&mut self, proposals: Vec<Proposal>) -> &mut Self {
```
danielSanchezQ, line 42: I don't know if this is the best approach for this. The builder pattern looks good. But, this does not let us modify the DB once it is built. I think some of the functionality could go in separate functions (that we can use for the builder too). But that way, will be more flexible in case we need to modifiy the DB mid test for example.

vit-servicing-station-tests/src/common/server.rs
```diff
@@ -0,0 +1,41 @@
...
+    process: Child,
+    settings: ServiceSettings,
+}
+
+impl Server {
+    pub fn new(process: Child, settings: ServiceSettings) -> Self {
+        Self { process, settings }
+    }
+
+    pub fn rest(&self) -> RestClient {
```
danielSanchezQ, line 15: I think `rest_client` would be better than plain `rest`?

vit-servicing-station-tests/src/common/server.rs
```diff
@@ -0,0 +1,41 @@
...
+        RestClient::new(self.settings.address.to_string())
+    }
+
+    pub fn rest_with_token(&self, hash: String) -> RestClient {
```
danielSanchezQ, line 19: Same-same here, `rest_client_with_token`?

vit-servicing-station-tests/src/common/data/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+use vit_servicing_station_lib::{
+    db::models::{
+        api_tokens::APITokenData,
+        proposals::{Category, Proposal, Proposer},
+        vote_options::VoteOptions,
+    },
+    v0::api_token::APIToken,
+};
+
+pub struct TestGen;
```
danielSanchezQ, line 11: There is no need to encapsulate this in a struct. Usually, you can have the same behavior simply with a `mod` (like namespacing) and public functions.

vit-servicing-station-tests/src/common/startup/server.rs
```diff
@@ -0,0 +1,71 @@
+use crate::common::startup::get_exe;
+use std::net::SocketAddr;
+use std::process::Command;
+use std::process::Stdio;
+use thiserror::Error;
+use vit_servicing_station_lib::server::settings::ServiceSettings;
+
+use crate::common::{paths::BLOCK0_BIN, server::Server, startup::get_available_port};
+
+pub struct Starter {
```
danielSanchezQ, line 10: Nitpicking: Can we use `ServerStarter`, `ServerBootstraper`, or something more specific here?

vit-servicing-station-tests/src/common/startup/mod.rs
```diff
@@ -10,7 +10,7 @@ use std::sync::atomic::{AtomicU32, Ordering};
 
 use crate::common::{data, server::Server};
 
-use self::{db::DbBuilder, server::Starter};
+use self::{db::DbBuilder, server::ServerBootstraper};
```
danielSanchezQ, line 13: MY BAD: it is **Bootstrapper**, sorry for that.

PR #44, opened 2020-07-07
[Tests] in_file_settings parameter bootstrap tests

dkijania:
New tests for server bootstrap:
- refactored code a little. extracted command builder and settings builder in order to gain more flexibility when defining settings and cmd parameters (including malformed paths etc.)
- added small pack of tests (which are going to be expanded in next PRs)


dkijania:
@danielSanchezQ  Cleaned up warnings

PR #45, opened 2020-07-07
Use health endpoint

dkijania:
Used health endpoint to verify server is up in automated tests

### DIFF
--- a/vit-servicing-station-tests/src/common/startup/mod.rs
+++ b/vit-servicing-station-tests/src/common/startup/mod.rs
@@ -11,6 +11,7 @@ use std::sync::atomic::{AtomicU32, Ordering};
 use crate::common::{data, server::Server};
 
 use self::{db::DbBuilder, server::ServerBootstrapper};
+use server::ServerBootstrapperError;
 
 pub fn get_exe() -> PathBuf {
     const VIT_BIN_NAME: &str = env!("VIT_BIN_NAME");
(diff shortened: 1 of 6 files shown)


vit-servicing-station-tests/src/common/startup/server.rs
```diff
@@ -39,7 +39,23 @@ impl ServerBootstrapper {
...
+
+    pub fn is_up(&self, server: &Server, token: &str) -> Result<(), ServerBootstrapperError> {
+        for _ in 0..3 {
+            if server.is_up(token.to_string()) {
+                return Ok(());
+            }
+            std::thread::sleep(std::time::Duration::from_secs(1));
+        }
+        Err(ServerBootstrapperError::FailToBootstrap)
+    }
```
danielSanchezQ, line 56: I think `start` should just start the server and should know nothing about the token. `is_up` can be called explicitly with the token in the test.

A following call would look like:
```rust
ServerBootstrapper::new()
...
.start().unwrap()
.is_up(token)
```

vit-servicing-station-lib/src/v0/endpoints/health/handlers.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        let result = warp::test::request()
+            .method("GET")
+            .path("/health")
+            .reply(&filter)
+            .await;
+
+        assert_eq!(result.status(), warp::http::StatusCode::OK);
+    }
+}
```
danielSanchezQ, line 35: These files shouldn't be in this PR. These changes were merged already. Maybe you need to rebase?

dkijania: yes, i was developing on top of your branch

vit-servicing-station-tests/src/common/startup/mod.rs
```diff
@@ -54,19 +55,21 @@ pub fn get_available_port() -> u32 {
...
 
     let server = ServerBootstrapper::new()
         .with_db_path(db_path.to_str().unwrap())
-        .start()
-        .unwrap();
+        .start()?;
 
-    (server, hash)
+    if !server.is_up(&hash) {
+        panic!("server is not up");
```
danielSanchezQ, line 71: Can we have an error for this?

dkijania: i've just converted it to error. we will see in next prs if check_running() would be useful. If yes i will added it

PR #51, opened 2020-07-09
Implemented application exit error codes

danielSanchezQ:
By using specific error codes makes the bootstrapping testing of the application cleaner and easier.

### DIFF
--- a/vit-servicing-station-server/src/main.rs
+++ b/vit-servicing-station-server/src/main.rs
@@ -13,7 +13,7 @@ async fn main() {
     if let Some(settings_file) = &settings.out_settings_file {
         server_settings::dump_settings_to_file(settings_file, &settings).unwrap_or_else(|e| {
             println!("Error writing settings to file {}: {}", settings_file, e);
-            std::process::exit(ApplicationExitCode::WriteSettingsError as i32)
+            std::process::exit(ApplicationExitCode::WriteSettingsError.into())
         });
         return;
     }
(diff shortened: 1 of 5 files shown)


vit-servicing-station-server/src/main.rs
```diff
@@ -12,7 +13,7 @@ async fn main() {
     if let Some(settings_file) = &settings.out_settings_file {
         server_settings::dump_settings_to_file(settings_file, &settings).unwrap_or_else(|e| {
             println!("Error writing settings to file {}: {}", settings_file, e);
-            std::process::exit(1)
+            std::process::exit(ApplicationExitCode::WriteSettingsError as i32)
```
dkijania, line 16: Maybe we can add `Into<i32> impl for ApplicationExitCode`

vit-servicing-station-tests/src/tests/bootstrap/parameters.rs
```diff
@@ -29,7 +35,8 @@ pub fn in_settings_file_does_not_exist() {
         .in_settings_file(&non_existing_file)
         .build()
         .assert()
-        .failure();
+        .failure()
+        .code(ApplicationExitCode::LoadSettingsError as i32);
```
danielSanchezQ, line 39: `into()` cannot be used here since the assert takes the type from the function parameter apparently. `Into<i32>` should be used, but for that case, it is more straight forward to use the `as` I think.

dkijania:
Looking good!

PR #52, opened 2020-07-10
[Tests] jortestkit submodule

dkijania:
to avoid code duplication between jormugandr and vit testing i introduced jortestkit project. It contians universal tools for testing. For example openssl which would be used for tls testing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 name: Continuous integration
 
 jobs:
+
   check:
     name: Check
     strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -12,6 +13,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v1
+      - uses: Mushus/checkout-submodule@v1
```
danielSanchezQ, line 16: I think in the checkout one you can ask for the submodules directly. How do we do it in jormungandr?

dkijania: will try with actions/checkout@v2

dkijania:
nope, doesn't work

danielSanchezQ:
Use the submodules statement:
```
- uses: actions/checkout@v2
  with:
    submodules: true
```

You can find some help [here](https://github.com/actions/checkout#usage)

dkijania:
thanks! that solves problem

PR #54, opened 2020-07-10
Use OutputJson for i64 graphql atributtes

danielSanchezQ:
This PR is a rollback: Async-graphql lib removed the possibility to use scalar as 64bit integers (forcing them to be encoded as JSON strings instead of numbers). But they introduced this new OutputJson that add another workaround for the problem.


danielSanchezQ:
This is a minor change and it pass the tests so I'm merging it.

PR #64, opened 2020-07-17
[Tests] test case with incorrect block0 and db_url

dkijania:


vit-servicing-station-tests/src/tests/bootstrap/parameters.rs
```diff
@@ -84,3 +88,23 @@ pub fn in_settings_file_with_malformed_path() {
...
+
+    let db_path = DbBuilder::new()
+        .with_token(token)
+        .with_proposals(data::proposals())
+        .build(&temp_dir)?;
+
+    let server = ServerBootstrapper::new()
+        .with_block0_path(db_path.to_str().unwrap())
+        .with_db_path(BLOCK0_BIN)
+        .start()?;
```
danielSanchezQ, line 106: The server will not be able to bootstrap now and will return the DB connection trouble exit code.

PR #66, opened 2020-07-17
[Tests] Graphql tests

dkijania:
New set of tests for graphQL test cases for `proposal`
New requests uses graphql requests:
- fields
- comments
- arguments
- aliases
- fragments
- invalid requests:
a) cycle
b) field does not exist

### DIFF
--- a/vit-servicing-station-tests/src/tests/graphql/proposals.rs
+++ b/vit-servicing-station-tests/src/tests/graphql/proposals.rs
@@ -122,7 +122,7 @@ pub fn test_graphql_requests() -> Result<(), Box<dyn std::error::Error>> {
 
     let graphql_client = server.graphql_client_with_token(&snapshot.token_hash());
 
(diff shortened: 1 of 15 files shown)


vit-servicing-station-tests/src/tests/graphql/proposals.rs
```diff
@@ -13,8 +119,26 @@ pub fn get_proposals_by_id() {
...
+
+    alias_test(&graphql_client, proposal_id);
+
+    cycle_test(&graphql_client, proposal_id);
+
+    proposal_without_arguments_test(&graphql_client);
+
+    proposals_test(&graphql_client, &snapshot);
+
+    proposal_with_fragments_test(&graphql_client, proposal_id);
```
danielSanchezQ, line 141: I think it would be better to have this tests completely separated, one in each `#[test]` case.

Issue #41, opened 2020-07-06
Introduce file logger


dkijania:
It would be nice to have an file logger. Currently user would need to read standard output of service to check for any panics/errors

danielSanchezQ:
Solved in #70

PR #70, opened 2020-07-23
Implemented file logging and logging configuration

danielSanchezQ:
Implements #41

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,7 +165,7 @@ dependencies = [
  "httparse",
  "indexmap",
  "itertools",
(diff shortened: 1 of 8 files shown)


danielSanchezQ:
@dkijania we would have to add some testing for the newly added configuration arguments.

danielSanchezQ:
I'm merging this. More things are coming on top of this small changes. Any fix can be done later as it is collateral to the main app.

dkijania:
Hi, sorry was testing it just now. looking good, so no objections with merging :)

PR #71, opened 2020-07-23
[Test] Funds graphql tests

dkijania:


vit-servicing-station-tests/src/tests/graphql/funds.rs
```diff
@@ -0,0 +1,105 @@
...
+                id: fund_id.to_string(),
+            }
+            .render()
+            .unwrap(),
+        )
+        .unwrap();
+    assert!(query_result["errors"][0]["message"]
+        .as_str()
+        .unwrap()
+        .contains("Invalid value for argument"));
```
danielSanchezQ, line 70: Can we check the error code too in this tests? I would prefer to rely on the codes than the messages (although we can keep the message checking).
Or are this the `graphql` errors? If they are standard errors derivated from `graphql` instead of our customized ones, then there is no really need of testing them (they should be standard...hopefully).

dkijania: Those are standard (i think) graphql errors. I can remove them.

danielSanchezQ: Leave them for now. But just knowing it should fail should be enough. 
If depending on the error messages that are not ours we may have to change them if they change the implementation of whatever lib we are using (or if we need to change lib).

PR #72, opened 2020-07-28
Logging lib

danielSanchezQ:
The idea behind this library is to have some structured logging messages that we can use to process data later. Some interesting things that we can get:

- Requests statistics
- User stats
- Crawlers detection
- API keys usage
...

For now, it just dumps the data into whatever log it is configured. It can be expanded later for dumping data into a DB or service for processing.


danielSanchezQ:
I want to integrate the new flags with the docker script. I'll open for review after #77 is merged.

dkijania (approved):

PR #73, opened 2020-07-29
Async graphql update i64

danielSanchezQ:
They implemented i64/u64 as a number instead of string when serializing objects so we can get rid of the previous wrapper.


danielSanchezQ:
I'm merging this small update/change. It was like this in the first versions of the service before they changed it in the `async_graphql` lib. Tests pass so it should be guarantee enough.

PR #74, opened 2020-07-30
[Tests] Soak test

dkijania:


vit-servicing-station-tests/src/common/data/generator.rs
```diff
@@ -47,12 +51,13 @@ impl Snapshot {
         self.funds.iter().find(|x| x.id == id)
     }
 
-    pub fn token(&self) -> (APITokenData, String) {
-        self.token.clone()
+    pub fn any_token(&self) -> (String, APITokenData) {
+        let (hash, token) = self.tokens.iter().next().clone().unwrap();
```
danielSanchezQ, line 55: I'm not sure if this is gonna work as expected. HashMaps are not ordered by default by I think the iteration will always run in the same order for any of them (we just don't know which order). So, if we want `any_token` to always return the same token for any `snapshot` then it is ok. If we want to randomly return a token for the ones available we would need to use some other implementation.

dkijania: My assumption was that it does not matter at this point since, we have no validation of token atm. in future i will intorduce any_valid_token method  or similar

vit-servicing-station-tests/src/common/db/mod.rs
```diff
@@ -16,17 +16,19 @@ impl<'a> DbInserter<'a> {
...
-            api_tokens::dsl::creation_time.eq(token_data.creation_time),
-            api_tokens::dsl::expire_time.eq(token_data.expire_time),
-        );
+    pub fn insert_tokens(&self, tokens_data: &[APITokenData]) -> Result<(), DbInserterError> {
+        for token_data in tokens_data {
+            let values = (
+                api_tokens::dsl::token.eq(token_data.token.as_ref()),
+                api_tokens::dsl::creation_time.eq(token_data.creation_time),
+                api_tokens::dsl::expire_time.eq(token_data.expire_time),
+            );
```
danielSanchezQ, line 25: `insert_tokens` could be another abstraction on what we had previously:

```rust
pub fn insert_token(&self, token_data: &APITokenData) -> Result<(), DbInserterError> {
        let values = (
            api_tokens::dsl::token.eq(token_data.token.as_ref()),
            api_tokens::dsl::creation_time.eq(token_data.creation_time),
            api_tokens::dsl::expire_time.eq(token_data.expire_time),
        );
  

        diesel::insert_into(api_tokens::table)
            .values(values)
            .execute(self.connection)
            .map_err(DbInserterError::DieselError)?;

        Ok(())
}
pub fn insert_tokens(&self, tokens_data: &[APITokenData]) -> Result<(), DbInserterError> {
    for token_data in tokens_data {
        insert_token(token_data)?;
    }
    Ok(())
}

```

So we can keep the method to insert a single token without wrapping it in a vec first.

vit-servicing-station-tests/src/tests/non_functional/mod.rs
```diff
@@ -1,53 +1,150 @@
...
+    fn random_proposal_id(&self) -> i32 {
+        let proposals = self.snapshot.proposals();
+        let random_idx = self.radnom_usize() % proposals.len();
+        proposals.get(random_idx).unwrap().internal_id.clone()
+    }
+    fn random_fund_id(&self) -> i32 {
+        let funds = self.snapshot.funds();
+        let random_idx = self.radnom_usize() % funds.len();
+        funds.get(random_idx).unwrap().id.clone()
+    }
```
danielSanchezQ, line 59: This feels like they can go somewhere else. It looks awkward that such random methods are within the `VitRestRequest`. I think they should go to a wrapper that holds the randomizer. `VitRestRequest` can own an instance of such wrapper and use the methods from it.

danielSanchezQ: Or maybe just make some functions in a mod.

vit-servicing-station-tests/src/tests/non_functional/mod.rs
```diff
@@ -1,53 +1,150 @@
...
 }
 
 impl Request for VitRestRequest {
     fn run(&self) -> Result<(), RequestFailure> {
-        match self.random.clone().next_u32() % 2 {
-            0 => self
-                .rest_client
+        let random_token = self.random_token();
```
danielSanchezQ, line 64: This would look like : `let random_token = RandomMod::random_token(self.snapshot.tokens())`

vit-servicing-station-tests/src/tests/non_functional/mod.rs
dkijania: Good idea, thanks

PR #75, opened 2020-07-30
[Tests] Logger tests

dkijania:


danielSanchezQ:
Uhm, I see some tests failing because of the new log cli argument. I'll take a deeper look.

vit-servicing-station-lib/src/server/settings/config.rs
```diff
@@ -280,6 +280,19 @@ impl FromStr for LogLevel {
...
+        match self {
+            Self::Disabled => write!(f, "disabled"),
+            Self::Error => write!(f, "error"),
+            Self::Warn => write!(f, "warn"),
+            Self::Info => write!(f, "info"),
+            Self::Debug => write!(f, "debug"),
+            Self::Trace => write!(f, "trace"),
+        }
+    }
+}
```
danielSanchezQ, line 294: This looks good, but what do you think of this idea?

Could we rely on the serde_json for this ? this would rarely change but just in case we add more levels (maybe some custom ones). 

```rust
impl fmt::Display for LogLevel {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
            write!(f, serde_json::to_string(&self).unwrap())
        }
    }
}
```

That way we can keep up the representation to match exactly how we serialize.

danielSanchezQ: Uhm, actually, thinking about this. It is clearer this way what we are doing, and it matches the FromStr implementation. So, never mind.

danielSanchezQ:
The failing test is working with #72 .

dkijania:
One of the tests is failing : 
`tests::bootstrap::arguments::out_settings::out_settings_file_from_cmdlin`:

```
  left: `ServiceSettings { in_settings_file: None, out_settings_file: None, address: V4(127.0.0.1:9113), tls: Tls { cert_file: None, priv_key_file: None }, cors: Cors { allowed_origins: None, max_age_secs: None }, db_url: "C:\\Users\\Dariusz\\AppData\\Local\\Temp\\.tmpP3R70D\\vit_station.db", block0_path: "../resources/tests/block0.bin", enable_api_tokens: false, log: Log { log_output_path: Some("C:\\Users\\Dariusz\\AppData\\Local\\Temp\\.tmpP3R70D\\logger.log"), log_level: Info } }`,
 right: `ServiceSettings { in_settings_file: None, out_settings_file: None, address: V4(127.0.0.1:9113), tls: Tls { cert_file: None, priv_key_file: None }, cors: Cors { allowed_origins: None, max_age_secs: None }, db_url: "C:\\Users\\Dariusz\\AppData\\Local\\Temp\\.tmpP3R70D\\vit_station.db", block0_path: "../resources/tests/block0.bin", enable_api_tokens: false, log: Log { log_output_path: None, log_level: Info } }`',
```

PR #81, opened 2020-08-19
Cli backup db file before operations

danielSanchezQ:
Diesel SQLite backend does not support some advanced features as updates. Nor it is easily configurable to use batching for cross tables operations. This solution should be enough for our needs (since we are not talking about huge database sizes). 
If at some point in the future we migrate to something like Postgress we could use some other approach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3256,6 +3256,7 @@ dependencies = [
  "rand 0.7.3",
  "serde",
  "structopt",
+ "tempfile",
  "vit-servicing-station-lib",
 ]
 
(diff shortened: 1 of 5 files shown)


rinor:
LGTM. If possible we could add some logs regarding the steps:
- back-up to file xxxx
- restore from file xxx (in case of error)

the clean is done from destructors in the case of `NamedTempFile` right?

rinor (approved):

danielSanchezQ:
@rinor IIRC The tmp file is removed by the OS when no handler is open for the descriptor. But let me confirm  this.

Issue #85, opened 2020-09-02
Server binary should react to OS signals


danielSanchezQ:
Right now there is no way of gracefully exit the server process. It should react to the OS signals and handle resources properly.

PR #86, opened 2020-09-02
Implemented signal handling for gracefully shutdown the server

danielSanchezQ:
Implements #85

### DIFF
--- a/vit-servicing-station-lib/src/server/signals.rs
+++ b/vit-servicing-station-lib/src/server/signals.rs
@@ -1,8 +1,8 @@
 use tokio::signal;
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(target_os = "unix")]
 pub async fn watch_signal_for_shutdown() {
-    let mut interrupt_stream = signal::unix::signal(signal::unix::SignalKind::interrupt())
+    let mut interrupt_signal = signal::unix::signal(signal::unix::SignalKind::interrupt())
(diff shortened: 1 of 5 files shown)


danielSanchezQ:
@dkijania 
It would be nice to have some simple tests for this at some point.

vit-servicing-station-lib/src/server/signals.rs
```diff
@@ -0,0 +1,25 @@
+use tokio::signal;
+
+#[cfg(not(target_os = "windows"))]
+pub async fn watch_signal_for_shutdown() {
+    let mut interrupt_stream = signal::unix::signal(signal::unix::SignalKind::interrupt())
```
mzabaluev, line 5: Should be named `interrupt_signal`, I guess?

vit-servicing-station-lib/src/server/signals.rs
```diff
@@ -0,0 +1,25 @@
+use tokio::signal;
+
+#[cfg(not(target_os = "windows"))]
```
mzabaluev, line 3: I think this should be the other way around: this function's body is dependent on target_os = "unix", whereas the Windows variant is actually cross-platform, so it should be a fallback for non-UNIX.

mzabaluev (changes requested):

mzabaluev (approved):

Issue #83, opened 2020-08-31
Additional data - fund/proposals endpoint


rinor:
Fund: 
- `voting_power_threshold` (lovelace)

Proposals:
- `proposal_impact_score` (0.0 - 5.0)
- `proposer_relevant_experience` (text)

PR #88, opened 2020-09-03
Added additional data fields to fund/proposal

danielSanchezQ:
Implements #83


rinor (changes requested):
The cli needs fixing/update to include the new fields since it fails loading the data from csv files

rinor (approved):

PR #96, opened 2020-10-18
vote encryption key - added

rinor:


vit-servicing-station-cli/src/csv/models.rs
```diff
@@ -79,6 +79,8 @@ pub struct Proposal {
     pub chain_committee_end_time: i64,
     #[serde(alias = "chainVoteplanPayload", default = "Default::default")]
     pub chain_voteplan_payload: String,
+    #[serde(alias = "chainVoteEncryptionKey", default = "Default::default")]
```
danielSanchezQ, line 82: Is it ok not to have the encryption key?

rinor: - for the `public` payload yes
- for the `private` one, we should have that key

but then again, not sure if this service should check/enforce it.

danielSanchezQ (approved):
LGTM

PR #104, opened 2020-11-20
Tracing

danielSanchezQ:
We decided it was better to migrate and use plain tracing instead of the old logging-lib.
This migrates the existing logs and previous configuration options.


mzabaluev (approved):

mzabaluev:
@danielSanchezQ I have rebased the branch and fixed up build for tests. There is one test that's failing due to an old config value present in the config file; I could not quickly find where is it coming from.

mzabaluev:
Hmm, the same test is passing in CI, there's only some Windows path ugliness to fix.
Do the tests persist some config files between runs?

danielSanchezQ:
I'm on it. One of them is because the appender library actually panics itself. So, we need to add a previous check on the path. I got the 3 of them reproduced in my machine. Shame on windows though.

PR #105, opened 2020-11-25
Convert jortestkit to git dependency

mzabaluev:
Also re-enabling tests, because submodules should not be a problem now.


danielSanchezQ:
@mzabaluev , just some lints there. But it should be ok. After that Ill merge

PR #106, opened 2020-12-11
Load fund id from CSV data

mzabaluev:
Fund ID was not used when loading the database from CSV.
Fix this by allowing non-zero ID to override the autoincrement column
value, keeping with the convention of using Default::default in serde
when the id is omitted.

Change the tests to use a distinct ID that would not be confused by
by the default autoincrement value assigned by SQLite.

FIXME: was unable to parameterize GraphQL queries in tests because the engine
says no.


danielSanchezQ:
I rebased latest changes. I'll be merging this as soon as tests show green.
Queries with variables will be addressed after updating the graphql-library.
I'll make dependabot start checking this repository too.

PR #107, opened 2020-12-15
Updated async-graphql to version 2

danielSanchezQ:


vit-servicing-station-tests/src/tests/graphql/proposals.rs
```diff
@@ -83,6 +83,7 @@ pub fn additional_field_test() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 #[test]
+#[ignore]
```
eugene-babichenko, line 86: Why ignore this test?

danielSanchezQ: It is checking some error reporting from the graphql lib itself. They changed the format(?) with the newest version. Hence we decided to disable it for no until @dkijania can take a deeper look at it.

eugene-babichenko (approved):

PR #110, opened 2021-01-07
CI improvements

mzabaluev:
- Cache cargo dependencies.
- Remove the jobs that run `cargo check`, mostly redundant work in the good case.
- Change triggering condition to only run tests on PRs and pushes to master.
- Use `actions-rs/clippy-check` for extra annotation goodness.


danielSanchezQ (approved):
LGTM

PR #113, opened 2021-01-13
add flake.nix

cleverca22:


flake.nix
```diff
@@ -0,0 +1,33 @@
+{
+  inputs = {
+    utils.url = "github:numtide/flake-utils";
```
manveru, line 3: Could you add nixpkgs as explicit input?

jonringer (approved):

PR #114, opened 2021-01-15
challenge API specification

eugene-babichenko:


doc/api/v0.yaml
```diff
@@ -175,3 +198,22 @@ components:
...
           format: int32
+
+    Challenge:
+      properties:
+        title:
+          type: string
+        description:
+          type: string
+        extended_description:
+          type: string
```
mzabaluev, line 209: Does this come from the IdeaScale data model?

doc/api/v0.yaml
```diff
@@ -175,3 +198,22 @@ components:
...
+        rewards_total:
+          type: integer
+          format: int32
```
mzabaluev, line 212: If this is in lovelace, should probably be int64 just to be sure?

eugene-babichenko: Yeah, sounds right.

doc/api/v0.yaml
eugene-babichenko: Mostly based on the UI mock-ups I was provided with. Need to clarify how do they agree with the IdeaScale.

vit-servicing-station-lib/src/db/models/challenges.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl Insertable<challenges::table> for Challenge {
+    type Values = (
+        diesel::dsl::Eq<challenges::id, i32>,
+        diesel::dsl::Eq<challenges::title, String>,
+        diesel::dsl::Eq<challenges::description, String>,
+        diesel::dsl::Eq<challenges::rewards_total, i64>,
+        diesel::dsl::Eq<challenges::fund_id, i32>,
+    );
```
danielSanchezQ, line 47: I think you can just derive this impls in this case.

danielSanchezQ (changes requested):
Check if the queryable and insertable implementation for `Challenge` can be derived. Everything else looks ok 😃

vit-servicing-station-lib/src/db/models/challenges.rs
eugene-babichenko: fair point

doc/api/v0.yaml
```diff
@@ -56,6 +68,35 @@ paths:
...
+  /api/v0/challenges/{id}:
+    get:
+      summary: Get challenges by id
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: integer
+          required: true
+      responses:
```
mzabaluev, line 92: Could also document a 404 status if the challenge is not found. We do it in at least four places in this spec, yay diligence.

vit-servicing-station-lib/migrations/2020-05-22-112032_initial/down.sql
```diff
@@ -3,5 +3,5 @@ DROP TABLE IF EXISTS proposals;
 DROP TABLE IF EXISTS funds;
 DROP TABLE IF EXISTS voteplans;
 DROP TABLE IF EXISTS api_tokens;
+DROP TABLE IF EXISTS challenges;
```
mzabaluev, line 6: I don't think modifying old migrations is a proper way to do it in diesel.
Migrations should be applied (or deapplied) sequentially and recorded in a meta-table, so any newly introduced change should be created as a new migration applying the delta to the effective schema after the previous migration..

vit-servicing-station-lib/migrations/2020-05-22-112032_initial/up.sql
```diff
@@ -34,7 +34,8 @@ create table proposals
     chain_proposal_id BLOB NOT NULL,
     chain_proposal_index BIGINT NOT NULL,
     chain_vote_options VARCHAR NOT NULL,
-    chain_voteplan_id VARCHAR NOT NULL
+    chain_voteplan_id VARCHAR NOT NULL,
+    challenge_id INTEGER NOT NULL
```
mzabaluev, line 38: Same applies here: it should be `alter table` in a new migration subdirectory, starting from templates created by the diesel CLI tool.

mzabaluev (changes requested):

vit-servicing-station-lib/migrations/2020-05-22-112032_initial/down.sql
danielSanchezQ: I think in this case its ok since we are using a different version from fund to fund.

eugene-babichenko: > I don't think modifying old migrations is a proper way to do it in diesel.

Agree

eugene-babichenko: This is actually problematic. Writing a proper `down.sql` script for the new migration requires `ALTER TABLE proposals DROP COLUMN challenge_id`, but SQLite don't support `DROP COLUMN`. You can copy the columns that should be left after a rollback, but that looks a bit odd.

mzabaluev: > You can copy the columns that should be left after a rollback, but that looks a bit odd.

It's a bit odd indeed, but the workaround [looks simple enough](https://stackoverflow.com/questions/5938048/delete-column-from-sqlite-table).

mzabaluev: > I think in this case its ok since we are using a different version from fund to fund.

True that, also our database is created once and never reused in different deployments.

Should we scrap the whole migration system for simplicity, then? Making changes to timestamped migrations looks like misuse to me. Maybe we should remove the timestamp cruft from the single migration directory's name and name it something prominently like `migrations/init`.

danielSanchezQ: We use the migration because the library helps in loading the scripts for testing purposes and for the CLI tool to initiate an empty DB. Mostly so we do not have to do it by hand or using some other script for that. 
We can check if we can rename it removing the timestamp and moving to another directory as you comment. It should be ok.

danielSanchezQ (commented):
LGTM

vit-servicing-station-lib/src/db/models/challenges.rs
danielSanchezQ: @eugene-babichenko they couldn't finally be derived?
```rust
#[derive(Queryable, Insertable)]
struct Challenge {
    ...
}
```

Maybe you need to add a tag for specifying the table for that.
```rust
#[derive(Insertable, Queryable)]
#[table_name = "challenges"]
```

danielSanchezQ (approved):

mzabaluev (approved):
I thought the migration directory names could do without misleading timestamps at all, but as long as this quirk is documented, it will do.

PR #132, opened 2021-02-10
Challenge type and Proposal polymorphism in the API

mzabaluev:
Define different proposal subtypes based on the value of the `challenge_type` field.

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -230,8 +230,8 @@ components:
       discriminator:
         propertyName: challenge_type
         mapping:
-          simple: SimpleProposal
-          community-choice: CommunityChoiceProposal
+          simple: "#/components/schemas/SimpleProposal"
+          community-choice: "#/components/schemas/CommunityChoiceProposal"
       allOf:
         - $ref: "#/components/schemas/Proposal"
         - type: object
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -7,6 +7,14 @@ info:
   contact:
     url: "http://github.com/input-output-hk/vit-servicing-station"
 
+tags:
+  - name: fund
+    description: Information on treasury fund campaigns.
+  - name: challenge
+    description: Information on challenges, structuring proposals within a fund.
+  - name: proposal
+    description: Information on funding proposals.
+
 servers:
   - url: "http://localhost"
 
@@ -17,7 +25,7 @@ paths:
       summary: Get available fund
       tags: [fund]
       description: |
-        Retrieves information on the current round of treasury fund voting.
+        Retrieves information on the current treasury fund campaign.
       responses:
         "200":
           description: Valid response
@@ -32,7 +40,7 @@ paths:
       summary: Get fund by id
       tags: [fund]
       description: |
-        Retrieves information on the identified round of treasury fund voting.
+        Retrieves information on the identified treasury fund campaign.
       parameters:
         - in: path
           name: id
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -51,7 +51,7 @@ paths:
             application/json:
               schema:
                 items:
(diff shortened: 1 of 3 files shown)


doc/api/v0.yaml
```diff
@@ -227,12 +227,47 @@ components:
         fund_id:
           type: integer
           format: int32
+        challenge_id:
+          type: integer
+          format: int32
+        challenge_type:
+          $ref: "#/components/schemas/ChallengeType"
```
mzabaluev, line 234: Perhaps it's better to group these in an object that shares part of the structure with `Challenge`, also adding the challenge title to facilitate use in an UI view.

mzabaluev: Or would be better, but then the discriminator thing will not work.

doc/api/v0.yaml
```diff
@@ -51,7 +51,7 @@ paths:
             application/json:
               schema:
                 items:
-                  $ref: "#/components/schemas/Proposal"
+                $ref: "#/components/schemas/ProposalWithChallengeInfo"
```
eugene-babichenko, line 54: Wrong indentation level.

doc/api/v0.yaml
```diff
@@ -224,15 +219,61 @@ components:
...
+        - $ref: "#/components/schemas/ProposalWithChallengeInfo"
+        - type: object
+          properties:
+            proposal_brief:
+              type: string
+            proposal_importance:
+              type: string
+            proposal_goal:
+              type: string
+            proposal_metrics:
```
eugene-babichenko, line 267: Is this also some textual description?

doc/api/v0.yaml
```diff
@@ -224,15 +219,61 @@ components:
...
+      type: string
+      enum:
+        - simple
+        - community-choice
+
+    ProposalWithChallengeInfo:
+      discriminator:
+        propertyName: challenge_type
+        mapping:
+          simple: SimpleProposal
```
eugene-babichenko, line 233: This should be a fully qualified path to the type (like in `$ref`).

eugene-babichenko (changes requested):
We need an automatic OpenAPI validator :)

doc/api/v0.yaml
mzabaluev: Yes, just another field describing the key metrics.
The cue is that all of these fields need proper OpenAPI descriptions.

doc/api/v0.yaml
mzabaluev: Do you think the example [in the spec](http://spec.openapis.org/oas/v3.0.3#discriminator-object) may be misleading?

eugene-babichenko: It is. The documentation on swagger.io provides an [example with fully-qualified path](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/).

doc/api/v0.yaml
mzabaluev: Good catch. Both levels are actually somewhat valid, so I guess this is why the extension I have for auto-formatting and linting OpenAPI did not complain.

doc/api/v0.yaml
mzabaluev: I actually think both are valid from the language of the spec, the swagger example just being extra pedantic.

mzabaluev:
> We need an automatic OpenAPI validator :)

Agree. Feel free to add a step to check this in the `api_gh` workflow.

.github/workflows/api_validate.yml
```diff
@@ -0,0 +1,20 @@
+name: API Check
```
danielSanchezQ, line 1: Looks like there are two jobs that validate the API

mzabaluev: Yeah, fixed that in a force-push.

eugene-babichenko (approved):

mzabaluev:
I got a little carried away with the linter, should be good now.

Issue #135, opened 2021-02-12
Improve internal implementation for Proposal polymorphism


mzabaluev:
Action points after #134:

- [x] Eliminate a bogus `challenge_type` member in the `Proposal` data model: #138
- [x] Switch from a wide sparse proposals table to challenge type-specific tables and unify the proposals data sets obtained from subqueries with joins specific to each challenge type. Alternatively, introduce a dynamic system of custom per-challenge field names to mirror IdeaScale (in harmony with the IdeaScale importer).

danielSanchezQ:
Implemented in https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9f4d4c1e76370d50e1e0722538ee4beb

PR #138, opened 2021-02-15
Remove challenge_type on proposals.csv loader and the data model

mzabaluev:
`proposals.csv` does not need to have a `challenge_type` column, it is redundant with the data taken from `challenges.csv`.

Bring the model types closer to the database tables they represent and also to the OpenAPI declaration. `FullProposalInfo` is a data model extension of `Proposal`.
GraphQL and its tests had to drop the `challengeType` field on `Proposal`. It's should be done in a better way in GraphQL, anyway: full challenge data mounted as a subquery on Proposal.

Part of #135


danielSanchezQ (approved):
LGTM

PR #141, opened 2021-02-18
Database schema improvements for proposal polymorphism

danielSanchezQ:
- [x] Changed database schema to remove unnecessary stored null fields
- [x] Create varying structure to hold the attributes
- [x] Adapt `FullProposalsInfo` to use the new structures
- [x] Adapt testing functionalities
- [x] Adapt tests
- [x] Add database insert statements for newly created tables
- [x] Adapt csv loader to insert newly structures


vit-servicing-station-cli/src/csv/models.rs
```diff
@@ -137,11 +140,48 @@ impl From<Proposal> for proposals::Proposal {
...
+                    proposal_solution,
+                }))
+            }
+            (
+                None,
+                Some(proposal_brief),
+                Some(proposal_importance),
+                Some(proposal_goal),
+                Some(proposal_metrics),
+            ) => Ok(ProposalChallengeInfo::Community(CommunityChallengeProposal {
```
danielSanchezQ, line 171: Since we don't have the information available of the type in the proposal itself we can match set attributes here.
It may be wiser to check against the challenge itself for data sanity, but It will also make it more of a heavy load computation. Should we ensure this, or do we trust data input enough to skip it?

vit-servicing-station-lib/migrations/2020-05-22-112032_setup_db/up.sql
```diff
@@ -33,8 +33,21 @@ create table proposals
...
     chain_vote_options VARCHAR NOT NULL,
     chain_voteplan_id VARCHAR NOT NULL,
-    challenge_id INTEGER NOT NULL,
-    proposal_solution VARCHAR,
+    challenge_id INTEGER NOT NULL
+--     proposal_solution VARCHAR,
+--     proposal_brief VARCHAR,
+--     proposal_importance VARCHAR,
+--     proposal_goal VARCHAR,
+--     proposal_metrics VARCHAR
```
danielSanchezQ, line 41: Leftovers. Will delete.

vit-servicing-station-lib/src/db/models/proposals.rs
```diff
@@ -250,11 +263,11 @@ impl Queryable<full_proposals_info::SqlType, DB> for Proposal {
...
-            proposal_solution: row.26,
-            proposal_brief: row.27,
-            proposal_importance: row.28,
-            proposal_goal: row.29,
-            proposal_metrics: row.30,
+            // proposal_solution: row.26,
+            // proposal_brief: row.27,
+            // proposal_importance: row.28,
+            // proposal_goal: row.29,
+            // proposal_metrics: row.30,
```
danielSanchezQ, line 270: To be removed

vit-servicing-station-cli/src/csv/models.rs
mzabaluev: I'd rather use challenge type from the challenge struct, if we can associate it at this point.

danielSanchezQ: We do not have a relationship build at that time. But we can build it using a hashmap or finding in the challenge array before this.

vit-servicing-station-lib/src/db/models/proposals.rs
```diff
@@ -109,46 +113,55 @@ pub struct Proposal {
...
 }
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
 pub struct FullProposalInfo {
     #[serde(flatten)]
     pub proposal: Proposal,
+    #[serde(flatten)]
+    pub challenge_info: ProposalChallengeInfo,
     #[serde(alias = "challengeType")]
     pub challenge_type: ChallengeType,
```
mzabaluev, line 139: In the ultimate perfect form of this, I'd like to eliminate the `challenge_type` member as well and synthesize the field from the `challenge_info` discriminant, but this will do for now.

vit-servicing-station-lib/src/db/models/proposals.rs
```diff
@@ -109,46 +113,55 @@ pub struct Proposal {
...
 }
 
+impl Serialize for ProposalChallengeInfo {
+    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
+        let serde_data: SerdeProposalChallengeInfo = self.clone().into();
```
mzabaluev, line 147: There is some proxy type form of `serde` attribute to avoid typing this explicitly, I think.

vit-servicing-station-lib/src/db/models/proposals.rs
```diff
@@ -321,74 +340,107 @@ impl Insertable<proposals::table> for Proposal {
...
-            proposals::proposal_brief.eq(self.proposal_brief),
-            proposals::proposal_importance.eq(self.proposal_importance),
-            proposals::proposal_goal.eq(self.proposal_goal),
-            proposals::proposal_metrics.eq(self.proposal_metrics),
         )
     }
 }
 
+impl TryFrom<SerdeProposalChallengeInfo> for ProposalChallengeInfo {
+    type Error = ();
```
mzabaluev, line 348: It's stylistically better to use a named unit type.

vit-servicing-station-tests/src/tests/graphql/proposals.rs
```diff
@@ -9,6 +9,7 @@ use askama::Template;
 use assert_fs::TempDir;
 
 #[test]
+#[ignore]
```
mzabaluev, line 12: So, no luck getting it pass?

vit-servicing-station-tests/src/tests/graphql/proposals.rs
```diff
@@ -172,6 +180,7 @@ pub fn proposal_without_arguments_test() -> Result<(), Box<dyn std::error::Error
 }
 
 #[test]
+#[ignore]
```
mzabaluev, line 183: As above

vit-servicing-station-lib/src/db/models/proposals.rs
danielSanchezQ: For us is ok. But for the API consumer, I don't think it is ok. Either we keep this and have the `ProposalChallengeInfo` data flattened. Or we make it nested and remove the `challenge_type`.

vit-servicing-station-tests/src/tests/graphql/proposals.rs
danielSanchezQ: I didn't try with these latest changes. Anyway, it was ignored because of the proposition of getting it removed, which is probably what I am gonna do after this one is merged. But I can get it a try for the completeness of the lib state.

vit-servicing-station-lib/src/db/models/proposals.rs
danielSanchezQ: `serde(remote)`, but in this case, it wouldn't work because it may fail to transform, we cannot implement a `From` safely for this.

vit-servicing-station-tests/src/tests/graphql/proposals.rs
mzabaluev: It's OK to delete the polymorphic fields from GraphQL and/or the tests, we can re-add them later and with proper fragment support in the schema if needed.

danielSanchezQ: Yeah, the issue is not doing or not the implementation. But to actually test it. Current serde implementation with `flatten+alias` does not work. If we are gonna drop the GraphQL support we can completely skip this.

mzabaluev (approved):

PR #154, opened 2021-04-12
Added proposer rewards to challenges for rewards automation

danielSanchezQ:


vit-servicing-station-lib/src/db/models/challenges.rs
```diff
@@ -30,9 +32,11 @@ impl Queryable<challenges::SqlType, Db> for Challenge {
         String,
         // 3 -> rewards_total
         i64,
-        // 4 -> fund_id
+        // 4 -> rewards_total
+        i64,
+        // 5 -> fund_id
         i32,
-        // 5 -> fund_url
+        // 6 -> fund_url
```
eugene-babichenko, line 39: Why not preserve the old order of these?

danielSanchezQ: Well, I tried to keep the rewards attributes close together. Just that.

danielSanchezQ: I just realize, the confusion may come because the 4# item was wrong it should have been `proposers_rewards`. I fixed it now.

eugene-babichenko (approved):

PR #158, opened 2021-04-15
Make service version part of server config

danielSanchezQ:


eugene-babichenko (approved):

andrcmdr (commented):
Reviewed. Code is well enough.

But we need to implement a local storage for stateful versioning in further.

PR #162, opened 2021-04-19
Rest API additions

mzabaluev:
* Added `registration_snapshot_time`, replacing `voting_power_info` which duplicates the value for backward compatibility.
* Represented `voting_power_threshold` in the API description, previously omitted.

### DIFF
--- a/vit-servicing-station-lib/src/db/models/funds.rs
+++ b/vit-servicing-station-lib/src/db/models/funds.rs
@@ -3,10 +3,12 @@ use crate::db::{
     schema::funds,
     Db,
 };
+use crate::utils::datetime::unix_timestamp_to_datetime;
 use diesel::{ExpressionMethods, Insertable, Queryable};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[serde(into = "FundWithLegacyFields")]
 pub struct Fund {
     #[serde(default = "Default::default")]
     pub id: i32,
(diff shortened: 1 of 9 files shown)


vit-servicing-station-lib/src/db/models/funds.rs
```diff
@@ -3,41 +3,82 @@ use crate::db::{
...
+    #[serde(deserialize_with = "crate::utils::serde::deserialize_unix_timestamp_from_rfc3339")]
+    pub registration_snapshot_time: i64,
     #[serde(alias = "chainVotePlans", default = "Vec::new")]
     pub chain_vote_plans: Vec<Voteplan>,
     #[serde(default = "Vec::new")]
     pub challenges: Vec<Challenge>,
 }
 
+#[derive(Serialize)]
+struct FundWithLegacyFields {
```
danielSanchezQ, line 42: It is not clear to me, why do we separate this now?

danielSanchezQ (approved):
LGTM

vit-servicing-station-lib/src/db/models/funds.rs
mzabaluev: It's the easiest way I found to present the fields in JSON, while synthesizing the legacy field from the value of `registration_snapshot_time`.

eugene-babichenko (approved):

eugene-babichenko:
I think the tests are needed to be updated as well, not sure. The test suite is failing. The implementation itself is looking good though.

mzabaluev:
WTF, I have fixed this problem in the force-push and the tests pass locally for me.

PR #164, opened 2021-04-21
Bump OpenAPI version and document object fields

mzabaluev:
Bump the OpenAPI version to reflect the changes in #162.
Document the fields on Fund, Proposal and VotePlan schema objects.
Left out challenges, these need some rework with update to OpenAPI 3.1.

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -143,70 +143,93 @@ components:
         id:
           type: integer
           format: int32
+          description: Identifier of the fund campaign.
         fund_name:
           type: string
+          description: Human-readable name of the fund campaign.
         fund_goal:
           type: string
+          description: Description of the campaign's goals.
         voting_power_info:
           type: string
           deprecated: true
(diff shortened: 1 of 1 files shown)


doc/api/v0.yaml
```diff
@@ -143,70 +143,93 @@ components:
...
+          description: Deprecated, same as registration_snapshot_time.
         voting_power_threshold:
           type: integer
           format: int64
+          description: |
+            Minimal amount of funds required for a valid voter registration.
+            This amount is in lovelace.
         rewards_info:
           type: string
+          # FIXME: document the purpose of this field or deprecate it
```
eugene-babichenko, line 165: Shall we leave it as it is? :) I'd prefer to not merge stuff with `FIXME` comments when possible.

mzabaluev: I preferred it to stick like a sore thumb so that someone some day fixes it, but OK.

eugene-babichenko (approved):

PR #167, opened 2021-06-14
log http api requests with zipkin headers

eugene-babichenko:


vit-servicing-station-lib/src/v0/mod.rs
```diff
@@ -16,9 +16,33 @@ pub async fn filter(
...
+            method = %info.method(),
+            path = info.path(),
+            version = ?info.version(),
+            remote_addr = Empty,
+            trace_id = Empty,
+            span_id = Empty,
+            parent_span_id = Empty,
+        );
+        if let Some(remote_addr) = info.remote_addr() {
+            span.record("remote_addr", &remote_addr.to_string().as_str());
```
mzabaluev, line 34: Is this going to be the address of the load balancer in our infrastructure?

mzabaluev (approved):

vit-servicing-station-lib/src/v0/mod.rs
danielSanchezQ: I think is the remote address making the request. It is saved in the `remote_addr` record attribute

danielSanchezQ (approved):
LGTM

vit-servicing-station-lib/src/v0/mod.rs
mzabaluev: I'm not sure this is what warp puts into the `info` structure. Or does it parse the headers set up by the proxy to get at the original source IP address?

PR #172, opened 2021-07-01
Ideascale importer

danielSanchezQ:
This is the first approach to an ideascale importer tool. Things need to be polished and I realize about some changes that will need to be done but are outside of the scope of this task (an issue is already open for it).

Still to do:

- [ ] Use correct proposal chain id
- [x] Add null option tag
- [x] Set rewards correctly
- [x] Cleanup


vit-servicing-station-cli/src/ideascale/mod.rs
```diff
@@ -0,0 +1,524 @@
...
+#[derive(Debug, StructOpt)]
+struct VotingParameters {
+    #[structopt(long, parse(try_from_str = chrono::DateTime::parse_from_rfc3339))]
+    vote_start_time: DateTime<FixedOffset>,
+    #[structopt(long, parse(try_from_str = chrono::DateTime::parse_from_rfc3339))]
+    vote_end_time: DateTime<FixedOffset>,
+    #[structopt(long, parse(try_from_str = chrono::DateTime::parse_from_rfc3339))]
+    vote_committee_time: DateTime<FixedOffset>,
+    chain_vote_encryption_key: String,
+}
```
danielSanchezQ, line 105: Should we extract this to a json file too?

danielSanchezQ:
Coded moved to catalyst toolbox.

PR #178, opened 2021-08-04
Reviews cli

danielSanchezQ:
Added load of advisor reviews via csv to the cli command.

Based on #177 , should be merged afterwards.


danielSanchezQ:
@dkijania Your changes lgtm

PR #182, opened 2021-08-18
Updated open api docs with advisor reviews

danielSanchezQ:


doc/api/v0.yaml
```diff
@@ -136,6 +138,27 @@ paths:
         "404":
           description: The requested challenge was not found
 
+  /api/v0/reviews/{proposal_id}:
```
mzabaluev, line 141: It's a bit weird that we're adding these `/v0` paths when we have a v1 API alongside.

Or was the versioning meant to apply piecemeal to the paths, `/v0/reviews` meaning "version 0 of the reviews endpoint"?

doc/api/v0.yaml
```diff
@@ -441,3 +464,40 @@ components:
...
+    ReviewTag:
+      type: string
+      enum:
+        - Alignment
+        - Verifiability
+        - Feasibility
+        - Impact
+        - Auditability
+
+    Review:
```
mzabaluev, line 477: Let's make the names consistent: `AdvisorReview` and its plural `AdvisorReviews`.
```suggestion
    AdvisorReview:
```

doc/api/v0.yaml
```diff
@@ -441,3 +464,40 @@ components:
...
+
+    AdvisorReviews:
+      allOf:
+        - $ref: "#/components/schemas/Review"
+        - type: object
+          properties:
+            proposals:
+              type: array
+              items:
+                $ref: "#/components/schemas/Review"
```
mzabaluev, line 503: I don't get it. How can this result object have properties from `Review` and also a `proposals` array with other `Review` objects?

danielSanchezQ: This is a copypasta error.

doc/api/v0.yaml
danielSanchezQ: It is meant to be the Api version.

mzabaluev: So it should be `/v1/reviews` then?

mzabaluev:
Please incorporate the changes for #183 as well.

doc/api/v0.yaml
danielSanchezQ: Well, we do not have a `v1` per se. We use v0 permanently (at least for now) since API it is changed anyway from fund to fund. And, we do not need to support api changes for now. Or should we?

doc/api/v0.yaml
```diff
@@ -441,3 +464,46 @@ components:
...
+      properties:
+        id:
+          type: integer
+          format: i32
+        proposal_id:
+          type: integer
+          format: i32
+        rating_given:
+          type: integer
+          format: i32
```
mzabaluev, line 487: It would be good to have a description field detailing the scale of the integer rating.
Also minimum and maximum values could be specced?

doc/api/v0.yaml
```diff
@@ -441,3 +464,46 @@ components:
...
+        [
+          {
+            "id": 1,
+            "proposal_id": 1234,
+            "rating_given": 0,
+            "assessor": "za_assessor_432",
+            "note": "foo bar",
+            "tag": "Alignment"
+          }
+        ]
```
mzabaluev, line 509: Missing newline.

mzabaluev (approved):

doc/api/v0.yaml
danielSanchezQ: yep, added now. Thanks :)

danielSanchezQ:
Updated with #184. Need to wait for it to be merged.

mzabaluev (approved):

PR #187, opened 2021-11-02
Fix OpenApi review ref

danielSanchezQ:


danielSanchezQ:
I'm self merging this. It just fixes a tiny line about references in the open api description.

PR #190, opened 2021-11-19
Add highlights field to challenges

danielSanchezQ:


vit-servicing-station-lib/migrations/2020-05-22-112032_setup_db/up.sql
```diff
@@ -79,7 +79,8 @@ create table challenges
     rewards_total BIGINT NOT NULL,
     proposers_rewards BIGINT NOT NULL,
     fund_id INTEGER NOT NULL,
-    challenge_url VARCHAR NOT NULL
+    challenge_url VARCHAR NOT NULL,
+    highlights VARCHAR
```
danielSanchezQ, line 83: Even if json is supported by sqlite as a field, Diesel do not support it as a type for now.
I took the happy path. This most probably will change in further Funds (in where we will include more metadata to the challenges). And will have to be done in a separated table (and maybe we can use the same view approach we use in other parts). 
For now it should be enough.

Mr-Leshiy (approved):
LGTM

dkijania (approved):

PR #193, opened 2021-11-29
Proposals voteplanid index search

danielSanchezQ:


vit-servicing-station-lib/src/v0/endpoints/proposals/logic.rs
```diff
@@ -15,3 +16,36 @@ pub async fn get_proposal(
...
+                    proposal_query.voteplan_id,
+                    proposal_query.indexes,
+                    pool.clone(),
+                ),
+            )
+        })
+        .collect();
+    let mut results = Vec::new();
+
+    for task in tasks {
```
enzoc4, line 39: did you consider `futures::future::join_all`? not that it would make much difference, though. I guess this usually would have only a few tasks, right?

vit-servicing-station-lib/src/v0/endpoints/proposals/requests.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ProposalVoteplanIdAndIndex {
```
enzoc4, line 4: I think we usually write `VotePlanId` in chain-libs, so I find the capitalization here a bit odd (for the `p`)

vit-servicing-station-lib/src/v0/endpoints/proposals/requests.rs
```diff
@@ -0,0 +1,9 @@
...
+    pub voteplan_id: String,
+    pub indexes: Vec<i64>,
```
enzoc4, line 6: why signed? Although I guess it's something related to the db.

doc/api/v0.yaml
```diff
@@ -58,6 +58,29 @@ paths:
...
+
+      responses:
+        "200":
+          description: Valid response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProposalWithChallengeInfo"
+        "404":
+          description: The requested proposals were not found
```
enzoc4, line 82: all of them? or any of them?

vit-servicing-station-lib/src/v0/endpoints/proposals/requests.rs
danielSanchezQ: Actually I need to check how is this returned in the API. Good point.

vit-servicing-station-lib/src/v0/endpoints/proposals/requests.rs
danielSanchezQ: Yep, db related (sadly).

doc/api/v0.yaml
danielSanchezQ: As per now it is any. Will update.

danielSanchezQ: Actually this was missplaced. If something is not found it will just not appear. Query may return an empty list.

enzoc4 (approved):

PR #194, opened 2021-12-09
functional tests for voteplan_id and index query

dkijania:


vit-servicing-station-tests/src/tests/rest/voteplan_id.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    assert!(actual_proposals.len() > 0);
+
+    for proposal in actual_proposals {
+        let expected_proposal = expected_proposals
+            .iter()
+            .find(|x| proposal.proposal.internal_id == x.internal_id)
+            .unwrap();
+
+        assert_eq!(&proposal.proposal, expected_proposal);
```
Zeegomo, line 47:
```suggestion
    expected_proposals.sort_by(|p| p.proposal_id);
    actual_proposals.sort_by(|p| p.proposal_id);
    assert_eq!(actual_proposals, expected_proposals);
```

vit-servicing-station-tests/src/common/data/mod.rs
```diff
@@ -42,3 +44,52 @@ pub fn challenges() -> Vec<Challenge> {
     let funds = gen.funds();
     gen.challenges(&funds)
 }
+
+pub fn valid_fund_snapshot() -> Snapshot {
```
Zeegomo, line 48: Would it be nicer to `impl Arbitrary for Snapshot`?

vit-servicing-station-tests/src/tests/rest/voteplan_id.rs
```diff
@@ -0,0 +1,49 @@
+use crate::common::data::valid_fund_snapshot;
+use crate::common::startup::{db::DbBuilder, server::ServerBootstrapper};
+use assert_fs::TempDir;
+use vit_servicing_station_lib::v0::endpoints::proposals::ProposalVoteplanIdAndIndexes;
+
+#[test]
+pub fn get_proposals_by_voteplan_id_and_index() {
+    let temp_dir = TempDir::new().unwrap().into_persistent();
+
+    let snapshot = valid_fund_snapshot();
```
Zeegomo, line 10: nit: would be nice to test behavior with multiple vote plans

vit-servicing-station-tests/src/common/clients/rest.rs
```diff
@@ -144,11 +145,33 @@ impl RestClient {
...
+    }
+
+    pub fn challenges(&self) -> Result<Vec<Challenge>, RestError> {
+        let response = self.challenges_raw()?;
+        self.verify_status_code(&response)?;
+        let content = response.text()?;
+        self.logger.log_text(&content);
+        serde_json::from_str(&content).map_err(RestError::CannotDeserialize)
+    }
+
```
Zeegomo, line 174: Is this unrelated?

vit-servicing-station-tests/src/common/clients/rest.rs
```diff
@@ -231,9 +258,7 @@ impl RestClient {
         self.origin = Some(origin.into());
     }
 
-    // this is kept as part of the client even if not used
-    #[allow(dead_code)]
-    fn post(&self, path: &str, data: String) -> Result<serde_json::Value, RestError> {
+    fn post(&self, path: &str, data: String) -> Result<String, RestError> {
         let client = reqwest::blocking::Client::new();
         let mut res = client.post(path).body(String::into_bytes(data.clone()));
```
Mr-Leshiy, line 263: is it needed `data.clone()` here ?

vit-servicing-station-tests/src/common/clients/rest.rs
```diff
@@ -144,11 +145,33 @@ impl RestClient {
         serde_json::from_str(&content).map_err(RestError::CannotDeserialize)
     }
 
+    pub fn proposals_by_voteplan_id_and_index(
+        &self,
+        request: ProposalsByVoteplanIdAndIndex,
```
Mr-Leshiy, line 150: maybe better to `request` by reference: `request: &ProposalsByVoteplanIdAndIndex`

vit-servicing-station-tests/src/tests/rest/voteplan_id.rs
dkijania: I think it won't work because actual_proposals and expected_proposals have different types

vit-servicing-station-tests/src/common/data/mod.rs
dkijania: added Arbitrary, however it implies use of quickcheck, so i've limited number of method invocation to 1. Please tell me if you like current setup of tests

vit-servicing-station-tests/src/common/clients/rest.rs
dkijania: oh it is, but i realized that in order to create integration tests in vit-testing on this commit i need to use more methods in rest .

vit-servicing-station-tests/src/common/clients/rest.rs
dkijania: removed

vit-servicing-station-tests/src/tests/rest/voteplan_id.rs
enzoc4: I guess his suggestion is to avoid the quadratic part. In that case you could do an extra loop on `expected_proposals.iter().zip(actual_proposals.iter())` and call `assert_eq` for each one, and it would be the same.

Or `expected_proposals` could be a `HashMap`.

I wonder if it really matters though.

dkijania: ok, right. implemented :)

Zeegomo: It was not just for inefficiency, the current method does not check that every expected proposal is in `actual_proposals`

vit-servicing-station-tests/src/common/data/mod.rs
Zeegomo: mmh, I kind of retract my suggestion now 😅 . I guess we already have testing at the lib level https://github.com/danielSanchezQ/vit-servicing-station/blob/b030c1a4ba6881a676d2b2991006af5a814c9478/vit-servicing-station-lib/src/v0/endpoints/proposals/handlers.rs#L106 and that would be the preferred place for property testing I guess, as it is a bit slow here. We can go with a simple test here for now and think of improving test coverage there

PR #196, opened 2022-01-25
Port catalyst fund7 to master

dkijania:
Port various small tests adjustments made in catalyst-fund7 branch to  master


dkijania:
@const-iohk can you double check flake version?

PR #197, opened 2022-02-21
add test coverage integration

dkijania:


.github/workflows/test.yml
```diff
@@ -136,3 +136,48 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: -- --deny warnings
+
+  test_coverage:
+    name: Test Coverage
```
Zeegomo, line 141: can we use cargo-tarpaulin here as well?

Zeegomo: oh, please ignore this, I overlooked the change

Zeegomo (approved):

PR #198, opened 2022-02-21
introduce voter groups within a single fund

eugene-babichenko:
This PR introduces the concept of voter groups to the API. The idea behind this is that you can have multiple groups of voters voting within a single fund. Members of a group use the same voting token to vote on a given set of proposals.

Ideas/questions for API refactors given this change:

- Is `chain_vote_plans` field in `Fund` ever used?
- Shall we remove a bunch of fields duplicating `VotePlan` fields in `Proposal` and just include `chain_vote_plan` field there?

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -57,12 +57,12 @@ paths:
         "404":
           description: The requested fund was not found
 
-  /api/v0/proposals:
+  /api/v0/proposals/{voter_group_id}:
     post:
       summary: Get proposal by id
       tags: [ proposal ]
       description: |
-        Retrieves queried proposals.
+        Retrieves queried proposals for the provided voter group identifier.
       requestBody:
         description: List of voteplan id and indexes query
         required: true
@@ -93,13 +93,14 @@ paths:
                 items:
                   $ref: "#/components/schemas/ProposalWithChallengeInfo"
 
-  /api/v0/proposals/{id}:
+  /api/v0/proposals/{id}/{voter_group_id}:
     get:
       operationId: getProposal
       summary: Get proposal by id
       tags: [proposal]
       description: |
-        Retrieves information on the identified proposal.
+        Retrieves information on the identified proposal for a voter with the
+        provided voter group id.
       parameters:
         - in: path
           name: id
@@ -171,6 +172,11 @@ paths:
           schema:
             type: integer
           required: true
+        - in: path
+          name: voter_group_id
+          schema:
+            type: integer
+          required: true
       responses:
         "200":
           description: Valid response
@@ -230,6 +236,10 @@ components:
           items:
             $ref: "#/components/schemas/VotePlan"
           description: Vote plans registered for voting in this fund campaign.
+        groups:
+          type: array
+          items:
+            $ref: "#/components/schemas/VoterGroup"
         challenges:
           type: array
           items:
@@ -266,6 +276,11 @@ components:
           type: integer
           format: int32
           description: The fund ID this vote plan belongs to.
+        voting_token:
+          type: string
+          format: [0-9a-f]{56}(\.[0-9a-f]{1,64})?
+          description: |
+            The identifier of voting power token used withing this plan.
 
     Proposal:
       properties:
@@ -559,3 +574,16 @@ components:
           items:
               type: integer
               format: i64
+
+    VoterGroup:
+      properties:
+        id: integer
+          description: Unique identifier of this voter group
+        name: string
+          description: Voter group name
+        voting_token:
+          type: string
+          format: [0-9a-f]{56}(\.[0-9a-f]{1,64})?
+          description: |
+            The identifier of voting power token used withing this group. All vote plans within a
+            group are guaranteed to use the same token.
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -62,7 +62,7 @@ paths:
       summary: Get proposal by id
       tags: [ proposal ]
       description: |
-        Retrieves queried proposals for the provided voter group identifier.
+        Retrieves queried proposals.
       requestBody:
         description: List of voteplan id and indexes query
         required: true
@@ -84,6 +84,13 @@ paths:
       tags: [proposal]
       description: |
         Lists all available proposals.
+      parameters:
+        - in: path
+          name: voter_group_id
+          description: Get proposals only for the specified vote group.
+          schema:
+            type: integer
+          required: false
       responses:
         "200":
           description: Valid response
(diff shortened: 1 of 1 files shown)


doc/api/v0.yaml
```diff
@@ -171,6 +172,11 @@ paths:
           schema:
             type: integer
           required: true
+        - in: path
+          name: voter_group_id
```
Zeegomo, line 176: isn't this missing in the path? and btw I don't why it should be required

doc/api/v0.yaml
```diff
@@ -230,6 +236,10 @@ components:
           items:
             $ref: "#/components/schemas/VotePlan"
           description: Vote plans registered for voting in this fund campaign.
+        groups:
```
Zeegomo, line 239: Should this be embedded on chain or injected by vit-ss?

doc/api/v0.yaml
```diff
@@ -93,13 +93,14 @@ paths:
                 items:
                   $ref: "#/components/schemas/ProposalWithChallengeInfo"
 
-  /api/v0/proposals/{id}:
+  /api/v0/proposals/{id}/{voter_group_id}:
```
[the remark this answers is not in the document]
eugene-babichenko: Oops, a leftover from the old idea.

doc/api/v0.yaml
eugene-babichenko: Yes, it is missing. Speaking of why should it be required: well, you've got a point, we may leave the possibility to get all proposals regardless of groups.

doc/api/v0.yaml
eugene-babichenko: Injected by vit-ss at this point.

doc/api/v0.yaml
```diff
@@ -57,12 +57,12 @@ paths:
...
           description: The requested fund was not found
 
-  /api/v0/proposals:
+  /api/v0/proposals/{voter_group_id}:
     post:
       summary: Get proposal by id
       tags: [ proposal ]
       description: |
-        Retrieves queried proposals.
+        Retrieves queried proposals for the provided voter group identifier.
```
Zeegomo, line 65: what about having two endpoints for:
* proposals by voteplan
* voteplans for voting group
Otherwise we might risk these two to be out of sync

doc/api/v0.yaml
```diff
@@ -559,3 +574,16 @@ components:
...
+        id: integer
+          description: Unique identifier of this voter group
+        name: string
+          description: Voter group name
+        voting_token:
+          type: string
+          format: [0-9a-f]{56}(\.[0-9a-f]{1,64})?
+          description: |
+            The identifier of voting power token used withing this group. All vote plans within a
+            group are guaranteed to use the same token.
```
Zeegomo, line 589: > All vote plans within a group are guaranteed to use the same token

Why does this matter?

eugene-babichenko: To simplify the actual UX.

1. You choose the group to use.
2. Within this group you have voting power `x`.
3. You see the proposals that you can vote for with this voting power.

doc/api/v0.yaml
eugene-babichenko: We're in a trap of using the same URL for `GET` and `POST`. For now I only specified `voter_group_id` to be only used in `GET`.

doc/api/v0.yaml
Zeegomo: Thinking of voting group as roles, I might be in the technical council and within that group participate in different elections (e.g. inter-technical team, everybody same voting power, and with the whole council, where maybe different people will be required to have different powers). IMHO it would be inconvenient for me as a user to select the right voting group each time when clearly they are related to the same role.

Zeegomo:
Should we add an endpoint to return to which voting groups a user belong for the given fund? Otherwise it might be difficult for the client to fetch this

Zeegomo:
If we want to go the way of "blockchain is truth" then what really matter are just voting tokens. The app could simply fetch the account state, look at available tokens and filter voteplans which allows to vote with that token. Nothing really needed besides maybe adding an endpoint that does the filtering for voting token on the server side.

All of the voter group stuff only exists outside of the chain, but it:
* Enhances the UX in the voting app. 
* Has real influence on rounds results.

The app could fetch some mapping between voting tokens and voter groups and display to the user "you belong to these Y groups according to info available. Here is a description of what these groups represent" (note there's no real meaning assigned to tokens on chain nor voting groups have special permissions or capabilities on chain, it's just a matter of conventions), and maybe let the user filter voteplans by voting group.

doc/api/v0.yaml
```diff
@@ -559,3 +575,16 @@ components:
           items:
               type: integer
               format: i64
+
+    VoterGroup:
+      properties:
+        id: integer
```
Zeegomo, line 581: wondering whether we could collate those two things and simply have a string identifier that could be used as a descriptive name as well, so that it's more versatile in other tools like cli utilities

doc/api/v0.yaml
Zeegomo: Btw, don't think we define vote plans in a voting group anyway so I don't know how to interpret the last sentence

enzoc4: > Btw, don't think we define vote plans in a voting group anyway so I don't know how to interpret the last sentence

Aren't those implicit?

doc/api/v0.yaml
```diff
@@ -266,6 +277,11 @@ components:
           type: integer
           format: int32
           description: The fund ID this vote plan belongs to.
+        voting_token:
```
enzoc4, line 280: shouldn't this be the `VoterGroup` or the group id?

Zeegomo: This is a direct representation of on-chain data AFAIK, could be useful if we want to be close to blockchain data (i.e. this endpoint could be served by a node as well if needed).

doc/api/v0.yaml
Zeegomo: I think they are implicitly defined by looking at what voteplan use the same token as the one of voting group X, which makes the statement a bit weird (i.e. could not be otherwise defined what are the vote plans in a voting groups)

doc/api/v0.yaml
enzoc4: > This is a direct representation of on-chain data AFAIK, could be useful if we want to be close to blockchain data (i.e. this endpoint could be served by a node as well if needed).

We have the fund_id there though, which I think doesn't have an on chain representation (although maybe I'm forgetting something here). Also, I only see this structure used in the `v0/fund/{id}` endpoint. It doesn't make much difference though, but considering that we don't have a way to map `voting token -> group` having the information there could be useful.

Zeegomo: > don't have a way to map voting token -> group having the information there could be useful.
We do, it's the reverse of groups -> voting tokens in the `/fund` endpoint (assuming it's bijective)

I'm slightly in favor of having the voting token here for two reasons:
* we don't have this information anywhere else
* it's "closer" to the blockchain
but having voter group could also work and maybe it's easier for the app

Mr-Leshiy:
> Members of a group use the same voting token to vote on a given set of proposals.

Am I right that the number of the voting groups is equal to number of different voting tokens ? 
So the generation of the new voting token, creates a new corresponding voter group ?

PR #200, opened 2022-02-23
voter groups: alternative proposal

eugene-babichenko:
Introduce 1-to-many relationship between proposal data and on-chain proposals. This enables users from different groups to vote on the same proposal while reducing data duplication in the database.

Built on top of https://github.com/input-output-hk/vit-servicing-station/pull/198

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -57,12 +57,12 @@ paths:
         "404":
           description: The requested fund was not found
 
-  /api/v0/proposals:
+  /api/v0/proposals/{voter_group_id}:
     post:
       summary: Get proposal by id
       tags: [ proposal ]
       description: |
-        Retrieves queried proposals.
+        Retrieves queried proposals for the provided voter group identifier.
       requestBody:
         description: List of voteplan id and indexes query
         required: true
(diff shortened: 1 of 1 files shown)


doc/api/v0.yaml
```diff
@@ -315,35 +361,11 @@ components:
...
-        chain_committee_end_time:
-          type: string
-          format: date-time
-          description: Date and time for the end of tallying on this proposal's vote plan.
-        chain_voteplan_payload:
-          type: string
-          description: |
-            Whether the voting is done using the public or the privacy-preserving protocol.
+        chain_info:
+          type: array
```
Zeegomo, line 365: So you would use voting groups as indexes? I think it can be quite inefficient if we deal with voting groups >> 0. A map would be better suited for this

doc/api/v0.yaml
```diff
@@ -559,3 +581,16 @@ components:
           items:
               type: integer
               format: i64
+
+    VoterGroup:
+      properties:
+        id: integer
```
Zeegomo, line 587: I think a string could offer more flexibility and possibly be more descriptive if we want to go that way

Zeegomo:
From a user perspective I think most of the time I'm only requesting proposals for one voting group at a time, so this doesn't really save much there. We are free to do this in the DB if we are interested, but that's not something that influence end users.
Any feedback from app team is welcomed

enzoc4:
> This enables users from different groups to vote on the same proposal 

I don't really see how does this enable that, each proposal still has a unique group. Is the idea to have _duplicated_ proposals on chain on different voteplans that then get tallied and merged?

> while reducing data duplication in the database.

But then I don't really get this part, does this refer to jormungandr's db or vitss's db? If the later, I don't think it matters that much, and I don't think the api should define the db schema anyway.

PR #201, opened 2022-02-23
Added template builder for static data.

dkijania:
We can use it in tests in order to specify input data for vit-servicing-station data

### DIFF
--- a/vit-servicing-station-tests/src/common/data/generator/voting/builder/mod.rs
+++ b/vit-servicing-station-tests/src/common/data/generator/voting/builder/mod.rs
@@ -24,14 +24,7 @@ impl ArbitraryValidVotePlanConfig {
         self
     }
 
-    pub fn challenge(mut self, mut challenge: ChallengeConfig) -> Self {
-        challenge
-            .proposals
-            .iter_mut()
-            .enumerate()
-            .for_each(|(i, mut p)| {
-                p.challenge_id = Some(i);
-            });
+    pub fn challenge(mut self, challenge: ChallengeConfig) -> Self {
         self.challenges.push(challenge);
         self
     }
--- a/vit-servicing-station-tests/src/common/data/generator/voting/builder/mod.rs
+++ b/vit-servicing-station-tests/src/common/data/generator/voting/builder/mod.rs
@@ -7,24 +7,24 @@ use crate::common::data::FundTemplate;
 use crate::common::data::ProposalTemplate;
 use crate::common::data::ReviewTemplate;
 use crate::common::data::ValidVotingTemplateGenerator;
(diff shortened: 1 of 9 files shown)


enzoc4 (approved):
I think this looks good. I find a bit strange that some builders are not really builders and something more like config holders, like `ProposalBuilder` and `ChallengeBuilder`, that don't even have a `build` function, so I find the naming and the usage a bit weird. 

That said, I'm not sure how I would change it, and I think it's probably nitpicking anyway.

dkijania:
Hmm, actually this is a good point, that those structs should not be named as builders but configs. Will try to come up with some better naming

vit-servicing-station-tests/src/common/data/generator/voting/builder/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+impl ArbitraryValidVotePlanBuilder {
+    pub fn challenges(mut self, challenges: Vec<ChallengeBuilder>) -> Self {
+        for challenge in challenges.into_iter() {
+            self = self.challenge(challenge);
+        }
+        self
+    }
+
+    pub fn challenge(mut self, mut challenge: ChallengeBuilder) -> Self {
+        challenge
```
Mr-Leshiy, line 28: I am wondering why it is needed modify `challenge` variable before pushing it into `self. challenges`, don't we expect that it should be correctly initialized outside this function, somewhere inside `ProposalBuilder` methods

dkijania: Very good point. My priority was to simplify config construction to maximum. Joining ids is tiresome process so wanted to move it to internals. My goal was that person who creates config only defines business parameters:

```
       let mut template = ArbitraryValidVotePlanConfig::default().challenge(
            ChallengeConfig::default()
                .rewards_total(1000)
                .proposers_rewards(1000)
                .proposal(ProposalConfig::default().funds(100)),
        );
```

Mr-Leshiy: Yes, I understood, but I think it will be better to move it inside `ProposalConfig::default()` function, or it is not in place of definition of the `default` function for the `ProposalConfig` ?

dkijania: I managed to move assigning challenges id using lazy static and proposals when adding them to challenge

Mr-Leshiy: Hmm, I think in this case previous implementation will be better, so lets revert it

Mr-Leshiy (approved):

PR #202, opened 2022-03-02
Dependency maintenance

Zeegomo:
* Switch from chrono to time-rs
* Fix clippy warnings
* Update dependencies to resolve sec advisories


vit-servicing-station-tests/src/common/data/generator/voting/template/arbitrary.rs
```diff
@@ -79,7 +79,7 @@ impl ArbitraryValidVotingTemplateGenerator {
             CompanyName()
                 .fake::<String>()
                 .to_lowercase()
-                .replace(" ", "-"),
+                .replace(' ', "-"),
```
enzoc4, line 82: Interesting that both work, I think this is the first time I see this being used

enzoc4 (approved):
LGTM

saibatizoku (approved):

PR #207, opened 2022-03-22
add snapshot endpoint

enzoc4:
the format of the snapshot file, and of the input key is still a
placeholder (but the latter I guess should be hex).

this adds a new endpoint at v0/snapshot/{voting_key} to retrieve the
voting power, a table to back it up, and a service that reloads this
table when changes are detected on the provided file.


doc/api/v0.yaml
```diff
@@ -179,6 +179,25 @@ paths:
               schema:
                 $ref: "#/components/schemas/AdvisorReviews"
 
+  /api/v0/snapshot/{voting_key}:
```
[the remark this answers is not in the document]
enzoc4: I'm not sure I follow, do you mean keeping two snapshots of the snapshot? Or all the history of snapshots? how do you specify which snapshot?

I think that during the voting round one could just hit the account endpoint in the node (basically what the app does already). That said, if we need more information, like for example the list of delegations of a user, then that wouldn't be possible anymore with the node and we would have to do it here.

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -0,0 +1,204 @@
...
+                    trace!(?event);
+
+                    match event.kind {
+                        EventKind::Modify(ModifyKind::Metadata(MetadataKind::WriteTime)) => {}
+                        EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any)) => {}
+                        EventKind::Create(CreateKind::File) => {}
+                        EventKind::Remove(RemoveKind::File) => {}
+                        EventKind::Access(AccessKind::Close(AccessMode::Write)) => {}
+                        EventKind::Modify(ModifyKind::Name(RenameMode::To)) => {}
+                        EventKind::Modify(ModifyKind::Name(RenameMode::Both))
```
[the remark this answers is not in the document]
enzoc4: I had that before, but I removed it mostly because I think it can lead to reading partial writes. I mean, ongoing writes.

When I discussed the task with Caslav he said the idea was to `rsync` the file, and on my machine that only triggers `ModifyKind::Name(Rename::To)` and `ModifyKind::Name(Rename::Both)`, so potentially we only need one of those.

The other ones I left mostly just in case.

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -0,0 +1,204 @@
...
+
+            Ok::<_, notify::Error>(watcher)
+        })
+        .instrument(span!(Level::INFO, "snapshot dir watcher initialization",))
+        .await
+        .unwrap()
+    }?;
+
+    {
+        let mut rx = rx.clone();
```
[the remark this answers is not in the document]
enzoc4: not needed, I was returning a clone before at some point, but I removed that, well spotted.

vit-servicing-station-lib/src/server/settings/default.rs
```diff
@@ -16,6 +16,7 @@ impl Default for ServiceSettings {
             enable_api_tokens: false,
             log: Log::default(),
             service_version: "".to_string(),
+            snapshot: "./resources/snapshot/snapshot.json".to_string(),
```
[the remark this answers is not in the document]
enzoc4: went ahead and changed the other fields to use the constants too, for the sake of consistency

doc/api/v0.yaml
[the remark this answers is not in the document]
Zeegomo: > how do you specify which snapshot?
Maybe by slot number on mainnet ? or date also, even though it's a bit less precise

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -0,0 +1,291 @@
...
+                    context = "failed to fill snapshot table from file",
+                    %error
+                );
+            }
+        }
+    }
+}
+
+#[tracing::instrument(skip(context))]
+pub async fn async_watch(path: PathBuf, context: SharedContext) -> Result<WatcherGuard, Error> {
```
Zeegomo, line 129: Wondering if this is the best way to handle updates. Seems to carry some additional limitations (need access to the instance filesystem, ) and complexity (e.g. watcher service, selecting interesting notifications, tag parsing, debouncer, ...) compared to a `PUT`/`POST` request  but still requires outside intervention to feed new data in.
What was the thought process behind this? I may be missing some considerations / aspects

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -0,0 +1,291 @@
...
+pub type VotingGroup = String;
+
+#[derive(Serialize, Deserialize)]
+pub struct VotingHIR {
+    pub voting_key: Identifier,
+    pub voting_group: VotingGroup,
+    pub voting_power: Value,
+}
+
+type RawSnapshot = Vec<VotingHIR>;
```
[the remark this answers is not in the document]
enzoc4: Well, I can change the name to `Snapshot` I guess. This part is probably going to change anyway later.

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -0,0 +1,291 @@
...
+                .values(
+                    snapshot
+                        .iter()
+                        .map(
+                            |VotingHIR {
+                                 voting_key,
+                                 voting_power,
+                                 ..
+                             }| SnapshotEntry {
+                                voting_key: voting_key.to_hex(),
```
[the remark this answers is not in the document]
enzoc4: Well, it is the primary key, so there is a check for that already. If we want to add the group then we can just extend the primary key to that row too.

vit-servicing-station-lib/src/server/snapshot_watcher.rs
enzoc4: Tagging @const-iohk here since reading from the fs was mostly his idea.

I think having a new endpoint where we POST the file content would be fine and simpler indeed. Or if you mean just posting something like "reaload this file and use this tag".

One question I have with that is whether we can restrict that endpoint to the public but use it ourselves (at the architecture, I know we could have an api key or something like that). 

Another consideration would be that if we need to restart the service then we need to remember re-posting too, since I think the db is restored in its initial state. But besides this, I don't have any problem with that.

Zeegomo: >One question I have with that is whether we can restrict that endpoint to the public but use it ourselves (at the architecture, I know we could have an api key or something like that).

In other places we have a proxy for public interaction that only allows some endpoints, maybe we can have something similar here

enzoc4: Another alternative, by the way, would be to provide a CLI tool/command that just writes to the DB directly. It would lock the db for a bit, but it should be pretty much insignificant, especially since for this it would be called infrequently. I'd imagine we can run processes on the instance.

vit-servicing-station-lib/src/server/snapshot_watcher.rs
```diff
@@ -237,16 +237,17 @@ async fn load_snapshot_table_from_file(
...
                                  voting_key,
                                  voting_power,
-                                 ..
+                                 voting_group,
                              }| SnapshotEntry {
                                 voting_key: voting_key.to_hex(),
-                                voting_power: u64::from(*voting_power) as i64,
+                                voting_power: u64::from(voting_power) as i64,
                                 tag: tag.clone(),
+                                voting_group: dbg!(voting_group),
```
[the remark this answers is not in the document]
enzoc4: you mean remove the `dbg!`? mostly I just forgot to do it

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        };
+
+        let key_len = key.len();
+
+        let mut result: Vec<(Group, Value)> = vec![];
+
+        for entries in entries.range(key.clone()..) {
+            let (k, v) = entries?;
+
+            if k[0..key_len] > key[..] {
```
Zeegomo, line 64: I guess it's because you encode groups in the key as well, but this is a bit cryptic without explanation.

Also it could panic if `k.len() < key_len`, but not sure if we want to check this

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+                break;
+            }
+
+            let group =
+                String::from_utf8(k[key_len..].to_vec()).map_err(|_| Error::InternalError)?;
```
Zeegomo, line 69: same panic behavior

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        };
+
+        for entry in snapshot.into_iter() {
+            let VoterHIR {
+                voting_key,
+                voting_group,
+                voting_power,
+            } = entry;
+
+            let mut key = Vec::with_capacity(4 + 32 + voting_group.as_bytes().len());
```
Zeegomo, line 159: those are a bit magic numbers

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        if let None = tag_id {
+            self.next_tag_id += 1;
+        }
+
+        Ok(())
+    }
+}
+
+pub fn new_context() -> Result<(SharedContext, UpdateHandle), Error> {
+    let db = sled::Config::new().temporary(true).open()?;
```
Zeegomo, line 205: could a disk-backed db be useful if we have to restart the service?

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        let entries = db.open_tree("entries")?;
+
+        let mut next_tag_id = None;
+
+        for kv in tags.iter() {
+            let (_, v) = kv?;
+
+            let id = u32::from_be_bytes(v.as_ref().try_into().unwrap());
+
+            next_tag_id = next_tag_id.map(|c| max(c, id)).or_else(|| Some(id));
```
Zeegomo, line 113: this could be rewritten as something like `tags.iter()[..].map(|(k, v)| u32::from_be_bytes(v.as_ref().try_into().unwrap())).max().unwrap_or(0)`

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+                })?;
+
+                Ok(())
+            })
+            .await
+            .unwrap()
+            .map_err(Error::DbTxError)?;
+        }
+
+        if let None = tag_id {
```
Zeegomo, line 196: we are checking this at least 3 times in this function, I think we can do this at the beginning and assumes we already have an entry for this tag in `tags`

snapshot-service/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+            let (tag, _) = entries?;
+            result.push(String::from_utf8(tag.to_vec()).map_err(|_| Error::InternalError)?);
+        }
+
+        Ok(result)
+    }
+}
+
+// do NOT implement/derive Clone for this. The implementation of update relies on &mut self and the
+// split in a reader type and a writer type is to enforce a single writer.
```
Zeegomo, line 93: wouldn't this be solved if the initial access to `tags` was inside the single db transaction?

Zeegomo (changes requested):
I think we are going in the good direction by splitting those services when needed.
Not sure this custom sled impl is actually easier than a diesel / SQLite table.

I would also like to remove the file watcher altogether, but we should sync with @const-iohk first

snapshot-service/src/lib.rs
enzoc4: it is disk-backed even with temporary, if it was in memory I would have used an normal std Hash/BTree map as I was using a few commits ago. 

And yes, but with the current implementation it will read all the files in the snapshots directory at bootstrap, so it's not really much of a difference except for saving some time. If we switch to an endpoint to update then it would be more useful.

snapshot-service/src/lib.rs
enzoc4: the problem is that the `TransactionalTree` doesn't have range queries. It's not that much for tags, it's just so that nothing gets in the middle of building the batch with the bunch of deletes and actually applying it.

enzoc4:
> Not sure this custom sled impl is actually easier than a diesel / SQLite table.

I think the code is more complicated, but it is simpler in the sense that it's just rust, there are no macros, no dsl, no need to deal with the fact that sqlite integer types are only signed. 

And besides, it's not like we need to do any joins, and the access patterns are pretty simple, the relational part of SQL doesn't really add that much.

That said, I'll change it to sqlite if you ask for that. Although I'd prefer to not use diesel anyway, and just use the sqlite crate directly, even if we wanted to switch to postgres, the queries here can be implemented with standard SQL I guess.

snapshot-service/src/lib.rs
Zeegomo: Correct, I was indeed thinking of the second option, but at least we've found some advantage of using the file based trigger

Zeegomo:
>That said, I'll change it to sqlite if you ask for that. Although I'd prefer to not use diesel anyway, and just use the sqlite crate directly, even if we wanted to switch to postgres, the queries here can be implemented with standard SQL I guess.

I do not care about whatever backend we use, just that we do not overly complicate the code. That being said, the current implementation is fine, especially if we can easily add persistence

snapshot-service/src/lib.rs
Zeegomo: But IIRC it should be quite easy to add persistence to this, right?

snapshot-service/src/lib.rs
```diff
@@ -156,30 +154,37 @@ impl UpdateHandle {
...
-                    if let None = tag_id {
-                        tags.insert(tag.as_bytes(), &next_tag_id)?;
+                (&tags, &entries, &seqs).transaction(move |(tags, entries, seqs)| {
+                    if let Tag::New(id) = &tag_id {
+                        tags.insert(tag.as_bytes(), id)?;
+                        seqs.insert(
+                            TAG_SEQ_KEY,
+                            &(TagId::from_be_bytes(id.as_ref().try_into().unwrap()) + 1)
+                                .to_be_bytes(),
+                        )?;
```
Zeegomo, line 187: Where do you remove existing keys in case of tag update?

Zeegomo (commented):
Just a note on something that I could have missed, otherwise looks good.

snapshot-service/src/lib.rs
enzoc4: If you are talking about removing the tag entry if the input vector is empty, I'm not doing it. But I'll add it in a bit. It probably makes more sense.

Zeegomo: I mean removing keys that are not present anymore in the new input (i.e. there's no way of removing keys once they're added). Could this be helpful?

Zeegomo (approved):

PR #215, opened 2022-03-25
add another build method for DbBuilder.

dkijania:
There is a need to build db not only to temporary folder, but also to arbitrary path. This is needed especially in vitup which is also using DbBuilder to create database file used for voting. Currently it is built to temporary folder and then it has to be copied to target folder. This pr will help to skip copy step and create database in final destination

### DIFF
--- a/vit-servicing-station-tests/src/common/startup/db.rs
+++ b/vit-servicing-station-tests/src/common/startup/db.rs
@@ -145,11 +145,12 @@ impl DbBuilder {
     }
 
     pub fn build(&self, temp_dir: &TempDir) -> Result<PathBuf, DbBuilderError> {
-        let db = temp_dir.child("vit_station.db");
-        let db_path = db
-            .path()
-            .to_str()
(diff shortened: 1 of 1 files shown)


vit-servicing-station-tests/src/common/startup/db.rs
```diff
@@ -145,11 +145,12 @@ impl DbBuilder {
     }
 
     pub fn build(&self, temp_dir: &TempDir) -> Result<PathBuf, DbBuilderError> {
-        let db = temp_dir.child("vit_station.db");
-        let db_path = db
-            .path()
-            .to_str()
-            .ok_or(DbBuilderError::CannotExtractTempPath)?;
+        self.build_into_path(temp_dir.child("vit_station.db").path())
```
Mr-Leshiy, line 148: What do you think to provide a const variable for the `"vit_station.db"` ?

Mr-Leshiy (approved):

PR #218, opened 2022-04-01
Backport nix GitHub workflow

garbas:
... and other Nix & github actions related commits from @garbas and @const-iohk


const-iohk (approved):
LGTM

PR #224, opened 2022-04-07
[CI] Bump version to 0.4.0 and fix release action

zeegomo:
Bump version and fix some problems in the release action that prevented it from completing successfully

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,9 +123,6 @@ jobs:
         cross: [false]
         include:
           # Windows
-          - config: {os: windows-latest, target: x86_64-pc-windows-gnu}
-            toolchain: stable-x86_64-pc-windows-gnu
-            cross: false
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -166,7 +163,9 @@ jobs:
           mkdir .cargo
           cat > .cargo/config.toml <<EOF
           [target.${{ matrix.config.target }}]
-          rustflags = ["-Ctarget-cpu=generic", "-Clto", "-Cembed-bitcode=yes"]
+          rustflags = ["-C", "target-cpu=generic"]
+          [profile.release]
+          lto = "thin"
```
enzoc4, line 168: we could also just add this in the `Cargo.toml`, is it better to just do it here?

enzoc4 (approved):

.github/workflows/release.yml
Zeegomo: Tbh I just picked the commit we had on jormungandr

PR #227, opened 2022-04-18
search api 

cameron1024:


vit-servicing-station-lib/src/db/queries/search/generic.rs
```diff
@@ -0,0 +1,46 @@
...
+    r2d2::{ConnectionManager, PooledConnection},
+    sql_types::Text,
+    Expression, TextExpressionMethods,
+};
+
+use crate::{db::DbConnection, v0::errors::HandleError};
+
+type SqlStr = Text;
+
+diesel::no_arg_sql_function!(RANDOM, (), "Represents the sql RANDOM() function");
```
enzoc4, line 17: Not sure how I feel about non-deterministic/pure queries, to be honest. But I don't see a way to set up a seed neither, without defining a custom function or a table with a random sequence, that is.

I'd imagine this is to reduce biases? is it really that necessary?

vit-servicing-station-lib/src/db/queries/search/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+            SearchSort::ProposalFunds => {
+                search(proposals, proposal_title, proposal_funds, query, conn)
+            }
+            SearchSort::ProposalAdvisor => {
+                search(proposals, proposal_title, proposer_name, query, conn)
+            }
+            SearchSort::ProposalTitle => {
+                search(proposals, proposal_title, proposal_title, query, conn)
+            }
+            _ => unreachable!(),
```
enzoc4, line 62: I would just move the error thrown inside `validate_table_column_sort` here and just get rid of that function and these `unreachable!`. Less risk of both functions getting out of sync in the future and reaching a panic. But it's a bit of a personal preference I guess.

cameron1024: Yeah this didn't make me feel great, part of me feels like there might be a different way of expressing this, perhaps something like:
```rust
enum SearchQuery {
  Challenges(ChallengeColumn),
  Proposals(ProposalColumn),
}

enum ChallengeColumn {
  Index,
  Title,
  // ...
}
// etc...
```
but I'm not sure how nice the json representation would be :shrug: 

But either way, if we stay with the current representation this seems like a sensible change :+1:

cameron1024: Perhaps this is better?
```rust
#[derive(Debug, Clone, Deserialize)]
pub struct Query {
    table: String,
    constraints: Vec<Constraint>,
}

#[derive(Debug, Clone, Deserialize)]
pub struct Constraint {
    column: String,
    kind: ConstraintKind,
}

#[derive(Debug, Clone, Deserialize)]
pub enum ConstraintKind {
    Contains(String),
    Range { from: i64, to: i64 },
}
```
This seems more flexible, not sure how hard it would be to implement, but I suspect it's doable. The json representation is also quite nice (with some tweaking)

vit-servicing-station-lib/src/db/queries/search/generic.rs
cameron1024: If I'm honest, I'm struggling to think up a use case for randomizing the results, it just said "random" on the ticket.

I guess because we currently don't support limits, the client could just randomize and we could drop this? A previous version of this did randomization in Rust, which is much easier to seed, but it still doesn't feel great, especially around scaling.

cameron1024:
Currently, all queries constraints are *AND*-ed together, it's an open question whether we want to be able to express more complex queries.

If so, it would probably be good to know to avoid breaking changes

vit-servicing-station-lib/src/db/queries/search/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+) -> Result<SearchResponse, HandleError> {
+    match table {
+        Table::Challenges => {
+            use crate::db::schema::challenges::dsl::*;
+            use Column::*;
+
+            let mut query = challenges.into_boxed();
+
+            for Constraint { search, column } in filter {
+                let search = format!("%{search}%");
```
Zeegomo, line 41: mmm this is user input, how well are we guarded against malicious inputs?

vit-servicing-station-lib/src/db/queries/search/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+        Table::Proposals => {
+            use crate::db::views_schema::full_proposals_info::dsl::*;
+            use full_proposals_info as proposals;
+            use Column::*;
+
+            let mut query = proposals.into_boxed();
+
+            for Constraint { search, column } in filter {
+                query = match column {
```
Zeegomo, line 75: I think we are missing the `%` here in search, or at least I would keep both endpoints consistent

Zeegomo (changes requested):

vit-servicing-station-lib/src/db/queries/search/mod.rs
cameron1024: I googled it when I write it, and I found something saying diesel does input sanitization, I'll see if I can find that comment again

cameron1024: https://github.com/diesel-rs/diesel/issues/229

cameron1024: Seems like it does, it's just not super well documented

saibatizoku (approved):

juan-cieri:
> Some changes in the latest commit:
> 
> * removed `RANDOM` and entire generic API
> * Queries are now boxed, adds runtime overhead but negligible compared to db call cost
> 
> The query format is now:
> 
> ```
> {
>   "table": "proposals",
>   "filter": [
>     { "search": "title contains this", "column": "title" },
>     { "search": "description contains this", "column": "desc" },
>   ],
> "order_by": [
>     {"column": "title"},
>     {"column": "desc", "descending": true} 
>   ]
> }
> ```
> 
> If no `filter`s are provided, it returns all the elements in that collection
> 
> If no `order_by`s are provided, it returns it in the order it gets it from the DB (i.e. by primary key).
> 
> A minimal query just looks like: `{ "table": "proposals" }`

@cameron1024 is it possible to provide some definitions of search capabilities in https://github.com/input-output-hk/vit-servicing-station/blob/master/doc/api/v0.yaml please? So we can browse and analyze all type of requests and responses.
Thanks!

cameron1024:
I've added some details to the yaml spec, but I'm quite new to openapi, so let me know if there's any other details/issues that need fixing

minikin:
@dkijania @Zeegomo I believe we can ignore CI failures here and merge this pr, right?

dkijania:
@minikin we cannot ignore test failure job: 
`CI / Test Suite (ubuntu-latest) (pull_request) Failing after 5m — Test Suite (ubuntu-latest)`

minikin:
@cameron1024 we need to fix `CI / Test Suite (ubuntu-latest) (pull_request)`, please.

cameron1024:
This one was my bad, I thought the only failure was the openAPI validator, didn't see the actual test failure. I'm working on a fix, hopefully by the end of the day

cameron1024:
Had to rerun tests a few times, seems flaky. I'm investigating but would be good to get this one in

PR #229, opened 2022-04-26
Voting groups api

enzoc4:
Supersedes #198 and #200 

I'm creating a different PR because I think it's a bit cleaner, since I didn't really implement the API as it is in neither of those. There are still a few things I could polish a bit more, and some more testing, but I'm opening it now because I think it's mostly ready for review.

# API

### proposals -> proposal

`GET v0/proposals/{id}` is now `GET v0/proposal/{id}/{group-id}`

I removed the `s` since it returns a single proposal, which I think is more consistent. Also added the group-id filter.

### get all proposals

`GET v0/proposals` is now `GET v0/proposals/{group-id}`

## Comments

`POST v0/proposals` is kept as it is.

# Data

Add new two tables

# proposals_voteplans

As was more of less hinted in #200 I added a 1-N relationship between ideascale proposals and on-chain proposals. I kept a unified view in the API though, because it's simpler.

This also adds a new csv to load this table. I think this approach is cleaner than duplicating each proposal but changing voteplan ids. The only cons I see with this is that as it is right now an ideascale proposal could be not associated with any on chain proposal if we mess up the construction, but we could add a check for that anyway.

# groups

Similar thing, this table maps tokens to groups names/ids. Also has its own new csv file to populate it. 

# Use-cases/flow

As it is right now, this is how I understand things should work.

Users can know which groups they belong to in two ways: 

One is that they fetch the account status from the node to get their voting tokens. Then fetch the fund with `v0/fund`, and filter by this (on client side)

Two is that they use the snapshot endpoint once it's finished, since it will provide the voting power for each group, along with the group id.

Then they can operate within one group at a time, by adding the parameter to the `proposal` endpoint. The rest should be equal to the way things work right now.


vit-servicing-station-lib/migrations/2020-05-22-112032_setup_db/up.sql
```diff
@@ -30,12 +30,18 @@ create table proposals
...
-    chain_proposal_index BIGINT NOT NULL,
     chain_vote_options VARCHAR NOT NULL,
-    chain_voteplan_id VARCHAR NOT NULL,
     challenge_id INTEGER NOT NULL
 );
 
+create table proposals_voteplans (
+    proposal_id VARCHAR NOT NULL primary key,
+    chain_voteplan_id VARCHAR NOT NULL,
+    chain_proposal_index BIGINT NOT NULL,
```
Zeegomo, line 40: does this really need to be bigint? I think we only support 256 proposals in a voteplan

Zeegomo (changes requested):

vit-servicing-station-lib/migrations/2020-05-22-112032_setup_db/up.sql
enzoc4: I guess not, I didn't really change it though, it was already a bigint. 

That said, I think it makes no difference in sqlite, since they say

```
INTEGER. The value is a signed integer, stored in 0, 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value.
```
and `BIGINT` has `INTEGER` affinity/storage class.

I could change it anyway though.

vit-servicing-station-lib/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license = "MIT OR Apache-2.0"
 async-trait = "0.1.33"
 base64 = "0.12.1"
 time = { version = "0.3", features = ["parsing", "formatting"] }
-diesel = { version = "1.4.5", features = ["sqlite", "r2d2"] }
+diesel = { version = "1.4.5", features = ["sqlite", "r2d2", "64-column-tables"] }
```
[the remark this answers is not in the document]
enzoc4: yeah, I wanted to avoid that... And started a refactor on the `Queryable` implementation on `FullProposalInfo` to avoid it, but it turned out to be a bit complicated and I threw it away. I'm not really super accustomed to diesel yet to be honest.

doc/api/v0.yaml
```diff
@@ -78,41 +78,56 @@ paths:
...
+          required: true
       responses:
         "200":
           description: Valid response
           content:
             application/json:
               schema:
                 items:
-                  $ref: "#/components/schemas/ProposalWithChallengeInfo"
+                  allOf:
```
Zeegomo, line 102: `allOf` is not needed

doc/api/v0.yaml
```diff
@@ -78,41 +78,56 @@ paths:
...
+            type: string
+          required: true
       responses:
         "200":
           description: Valid response
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/ProposalWithChallengeInfo"
+                allOf:
```
Zeegomo, line 129: same

Zeegomo (approved):
Looks good, just a few nits to fix

PR #231, opened 2022-05-03
Multiple fund and multiple stage dates

enzoc4:
Still may need some tweaking/refactoring here and there.

This PR

Removes the restriction on only having a single fund, instead it assumes that the first one in csv is the current one.

Adds a getter endpoint for all the funds, in the order they were inserted.

Adds a PUT endpoint at `v0/fund` that allows editing/replacing the funds (or creating a new one, but maybe this should be disabled). I'm still unsure on whether this needs some sort of authentication or not.

Adds some fields to the `Fund` model, for dates of different stages.


Zeegomo (changes requested):

doc/api/v0.yaml
```diff
@@ -56,6 +56,45 @@ paths:
...
+        "422":
+          description: The input is malformed.
+
+  /api/v0/funds/:
+    get:
+      operationId: getFunds
+      summary: Get list of the funds
+      tags: [fund]
+      description: |
+        Get list of all the funds in the db. Ordered in time, which means the first one is the current one.
```
Zeegomo, line 87: maybe it's enough to return all the funds ids? I'm not sure there are a lot of cases where you want to fetch all of them

enzoc4: Well, it is mostly to save a request, since this would be queried all the time.

And I was expecting that we just add the two that we care about for now, we can add pagination arguments in the future (like `start=current&limit=2`) if necessary.

Another option would be to just return the current and the next in `v0/fund`, since that endpoint is a bit strange anyway, and is the one that the app currently uses, and keep this only for ids.

Zeegomo: My understanding is that we want to keep old entries as well, how do we handle those?

>which means the first one is the current one.

like this would not be true anymore I think

enzoc4: If we add old funds, yes, but are we going to do that? 

Currently you can just have the current one. And even here I only removed the restriction on having one, but I'm still using the first one as the current (and `v0/funds` does the same).

But I can remove that part of the comment for it to be more generic.

Zeegomo: Depends on requirements, when I read `continuous` I interpreted it as not restarting the instance each fund but rather add new stuff, which would mean that we return old stuff as well. But if that's not needed we can just restart it without any past data

vit-servicing-station-lib/src/db/queries/funds.rs
```diff
@@ -40,43 +45,66 @@ pub async fn query_fund_by_id(id: i32, pool: &DbConnectionPool) -> Result<Fund,
...
+}
+
+pub async fn query_current_fund(pool: &DbConnectionPool) -> Result<FundWithNext, HandleError> {
     let db_conn = pool.get().map_err(HandleError::DatabaseError)?;
     tokio::task::spawn_blocking(move || {
-        let fund = fund_dsl::funds
+        let funds: Vec<Fund> = fund_dsl::funds
             // TODO: Not sure if sorting by the PK is actually necessary
+            //
+            // this assumes that the next will be the second inserted
```
Zeegomo, line 70: we should make it clear in the interface as well

Zeegomo (approved):

doc/api/v0.yaml
```diff
@@ -56,6 +58,41 @@ paths:
...
+          required: true
+      responses:
+        "200":
+          description: Valid response
+        "422":
+          description: The input is malformed.
+
+  /api/v0/funds/:
+    get:
+      operationId: getFunds
```
Zeegomo, line 81: should this return just the ids now?

vit-servicing-station-lib/src/db/queries/funds.rs
enzoc4: which interface, exactly?

doc/api/v0.yaml
enzoc4: done, I wonder why is the CI not running though

vit-servicing-station-lib/src/db/queries/funds.rs
Zeegomo: the one we use to insert new funds, but not sure we have that currently :)

enzoc4: I mean, it's inserted from a csv... I guess I could add a comment in that argument's documentation. At some point I considered validating that the fund start dates were in order, but then I changed my mind.

PR #236, opened 2022-05-17
Add funds goals

enzoc4:
add goals table

an open question would be what should we do with the `fund_goal` field in `Fund`, since this makes it redundant (I guess?)

### DIFF
--- a/vit-servicing-station-cli/src/csv/loaders.rs
+++ b/vit-servicing-station-cli/src/csv/loaders.rs
@@ -8,6 +8,7 @@ use std::io;
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 use thiserror::Error;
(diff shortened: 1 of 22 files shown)


vit-servicing-station-cli/src/csv/loaders.rs
```diff
@@ -56,6 +57,10 @@ pub enum CsvDataCmd {
         /// Path to the csv containing advisor reviews information
         #[structopt(long = "reviews")]
         reviews: PathBuf,
+
+        /// Path to the csv containing advisor reviews information
```
Zeegomo, line 61: outdated comment

Zeegomo (approved):
>an open question would be what should we do with the fund_goal field in Fund, since this makes it redundant (I guess?)

Question for @juan-cieri

PR #249, opened 2022-07-11
NPG-2130 Add "native" ChallengeType to Challenges.

coire1:
This change will allow the endpoint `/api/v0/challenges` to expose an addition value `native` for the
`challenge_type` field and enable filtering in the voting app.


minikin:
@coire1 can you fix CI, pls?

PR #250, opened 2022-07-11
Update search api, add pagination functionality

Mr-Leshiy:
-  modify `/api/v0/search` request type, extended it with the new optional fields `limit` and `offset`, which can be used to limit the query result and provide an offset of the resulted data
```
    SearchQuery:
      properties:
        table:
          $ref: "#/components/schemas/SearchConstraint"
        filter:
          type: array
          items: 
            $ref: "#/components/schemas/SearchConstraint"
        order-by:
          type: array
          items: 
            $ref: "#/components/schemas/SearchOrderBy"
        limit:
          type: integer
          format: i32
          description: Sets the limit clause of the search query.
        offset:
          type: integer
          format: i32
          description: Sets the offset clause of the search query.
      required: [table]
```
- added new endpoint `/api/v0/search_count` which returns a count of the query results. Response type is the oldest response type for `/api/v0/search`. Actually it is quite the same if you will execute `/api/v0/search().len()`, but does it in more performance way.
```
    SearchCountQuery:
      properties:
        table:
          $ref: "#/components/schemas/SearchConstraint"
        filter:
          type: array
          items: 
            $ref: "#/components/schemas/SearchConstraint"
        order-by:
          type: array
          items: 
            $ref: "#/components/schemas/SearchOrderBy"
      required: [table]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,7 +464,7 @@ name = "chain-vote"
 version = "0.1.0"
 source = "git+https://github.com/input-output-hk/chain-libs.git?branch=master#ef798477aba3794c7b4b171f0926884bc0327563"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
  "chain-core",
  "chain-crypto",
  "const_format",
@@ -798,9 +798,9 @@ dependencies = [
 
 [[package]]
 name = "diesel"
-version = "1.4.6"
+version = "1.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "047bfc4d5c3bd2ef6ca6f981941046113524b9a9f9a7cbdfdd7ff40f58e6f542"
+checksum = "b28135ecf6b7d446b43e27e225622a038cc4e2930a1022f51cdb97ada19b8e4d"
 dependencies = [
  "byteorder",
  "diesel_derives",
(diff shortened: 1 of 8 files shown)


cameron1024 (commented):
Just a question about the general direction: what's the thinking behind splitting out the structs `SearchQuery` and `SearchCountQuery`? I'm a little worried about the duplication of the query code between `fn search` and `fn search_count`.

If it's not possible to join them into a single struct, we could at least pull the query-generating code into its own `fn get_query(SearchCountQuery) -> BoxedQuery` and reuse it between the two functions, but it might be simpler to just have a single `SearchQuery` struct with the extra pagination fields added?

Though maybe I'm missing something? Is there a reason why that doesn't work?

Mr-Leshiy:
@cameron1024 thanks for your suggestions !
> Just a question about the general direction: what's the thinking behind splitting out the structs SearchQuery and SearchCountQuery? I'm a little worried about the duplication of the query code between fn search and fn search_count

I don't want to say that it is impossible, but I am not sure that it is something that we really want. The reason in the returning type differences between `search` and `search_count`, at one place we are returning Vec and in another `i64`. 

> If it's not possible to join them into a single struct, we could at least pull the query-generating code into its own fn get_query(SearchCountQuery) -> BoxedQuery and reuse it between the two functions,

I am agree to follow such approach. Will try do it.

cameron1024:
Got it, didn't see the return types, that makes sense 😁

cameron1024 (approved):
LGTM :tada: 

It's a bit of a shame we're not using `derive(Insertable)` on our data types,it makes some of the type annotations a bit big, but that's a tech debt issue, nothing to do with this PR :grin:

PR #272, opened 2022-09-12
Define an API for the new /api/v0/user/{id} endpoint

Mr-Leshiy:
`/api/v0/user/{id}` should provide a needed information about the provided Cardano user id. User's information at the first glance should contain the following data:
- is_voter - flag which shows is this user a representative and actually can participant at this voting or it is a delegator.
- dreps -  list of delegated representatives for the current user
- voting_groups - voting group ids (voting tokens ids) at which current user participating or any of his representative take a part.

### DIFF
--- a/doc/api/v0.yaml
+++ b/doc/api/v0.yaml
@@ -1117,7 +1117,7 @@ components:
           items: 
             description: Voting group id (the same as voting token id)
             type: string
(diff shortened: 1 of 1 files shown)


stevenj (approved):
Looks good to me.

PR #273, opened 2022-09-14
Move all snapshot-service code into the vit-servicing-station-lib, remove snapshot-service crate

Mr-Leshiy:
It is basic cleanup, so I have moved all the implementation of the snapshot related endpoints into the vit-servicing-station-lib where actually service is initializing and implementation of other endpoints takes place


vit-servicing-station-lib/src/v0/endpoints/snapshot/routes.rs
```diff
@@ -9,17 +11,36 @@ pub fn filter(
...
-        .and_then(get_tags)
-        .boxed();
+        .and_then(get_tags);
 
-    root.and(get_voting_power.or(get_tags)).boxed()
+    root.and(get_voters_info.or(get_tags))
+}
+
+pub fn update_filter(
+    context: UpdateHandle,
```
cong-or, line 28: where is the `/api/v0/user/{id}` endpoint implemented ?

This seems to be snapshot related rather than user related - re `3367` ?

Mr-Leshiy: This pr contains only refactor, I have not done any modification or update here. I have only moved code

cong-or (approved):

